PR #5, opened 2018-06-28
Add --cleanup option

### LINKED ISSUE #4: Remove dev packages after building
This script installs a lot of packages that are needed to build the php extensions. It would be nice that it also removes the ones that are not needed afterwards in order to keep layers small.

These are mostly the ones that end with `*-dev`.

mlocati:
Fix #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ RUN chmod uga+x /usr/local/bin/install-php-extensions && sync && \
     install-php-extensions gd xdebug
 ```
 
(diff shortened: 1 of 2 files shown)


Hipska:
I would name it `--cleanup` if you later want to also support Alpine version..

Hipska:
Looks good to me 👍

mlocati:
Ok, let's merge it, and let's wait for feedback about potential problems :wink:

PR #22, opened 2019-07-13
Deprecated zip extension install warning

StefanPrintezis:
Use of bundled libzip is deprecated and will be removed.


Hipska:
So with this you introduce install warning?

mlocati:
I'm on holidays right now, a few thousand miles away from my PC.
Did someone test this change?

mlocati:
Superseded by #24

PR #58, opened 2019-12-16
Add support for grpc and protobuf

mcfedr:
I've moved to using your (excellent) script for installing my modules, and wanted to add a couple more that I can using.

Test: gprc, opencensus, protobuf


mlocati:
Great, thanks!
I guess we should always add `$PHPIZE_DEPS` to `$buildRequiredPackageLists_volatile` for Alpine: it's almost always required.

PS: if you add new extensions to `data/supported-extensions` file, you don't have to add `Test: ...` to the commit message

mcfedr:
Yes, just seen the build fail for PHPIZE, updated now.

mcfedr:
And now there is an issue I'm less sure on how to fix. That grpc module is producing too much output for travis.

mlocati:
> That grpc module is producing too much output for travis.

I tried to build grpc on Alpine with PHP 5.6, and it worked flawlessly. I think the issue is only that the output of the command is too long.
I'm looking at the PEAR source code, to see if there's a way to add `--quiet --enable-silent-rules` to the `configure` command, and `-silent` to the `make` command...

mlocati:
It seems that opencensus requires PHP 7.0+

mcfedr:
> It seems that opencensus requires PHP 7.0+

Yes, I've removed it for the moment, it wasnt building correctly - I think even on 7 - in my current image I am building it from source, not from pecl, and in trying this out I remember why, there are some fixes that are currently not in a release.

mlocati:
> Yes, I've removed it for the moment,

Did you remove it from where?

> it wasnt building correctly - I think even on 7

Yep: since its very first release, [it requires PHP 7.0+](https://github.com/census-instrumentation/opencensus-php/blob/v0.0.4/ext/package.xml#L98).

We also have [this error](https://github.com/census-instrumentation/opencensus-php/issues/235) in [PHP 7.3](https://travis-ci.org/mlocati/docker-php-extension-installer/jobs/625760846#L516) and in [PHP 7.4](https://travis-ci.org/mlocati/docker-php-extension-installer/jobs/625760846#L763)

So, for the time being, I'd add support only to PHP 7.0, 7.1 and 7.2

mcfedr:
Ah, looks like it got lost in your fixes, well if it builds in those versions might as well leave it

I pushed in 8213f41.

mlocati:
> Ah, looks like it got lost in your fixes

Whoops, sorry :wink:

> if it builds in those versions might as well leave it

The test script is still trying to check opencensus because it finds a commit containing `Test: opencensus`.
I'm going to fix this.

mlocati:
For opencensus I created (and merged) #60

mlocati:
:+1:

mcfedr:
Awesome, thanks for creating this tool!

mlocati:
> Awesome, thanks for creating this tool!

:wink: I was quite tired to look for instructions describing how to install every extension, and I'm glad that it's handy for others too.

PR #59, opened 2019-12-16
Simpler add/remove packages

mcfedr:
This is more an idea that a ready to go PR, and its possible you have done something similar and have good reasons for doing it differently, but I have been trying this out and it seems to work and I thought I'd share it.

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -177,8 +177,8 @@ sortModulesToInstall () {
 #   0 (true): if we to install some package
 #   1 (false): if no package is required
 buildRequiredPackageLists () {
-	buildRequiredPackageLists_persistent=''
-	buildRequiredPackageLists_volatile=''
+	PACKAGES_PERSISTENT=''
+	PACKAGES_VOLATILE=''
 	buildRequiredPackageLists_distro="$(getDistro)"
 	buildRequiredPackageLists_phpv=$1
 	while :; do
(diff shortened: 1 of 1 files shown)


install-php-extensions
```diff
@@ -550,12 +550,10 @@ buildRequiredPackageLists () {
...
 			;;
 	esac
-	if test -n "$buildRequiredPackageLists_persistent"; then
-		PACKAGES_PERSISTENT="$(expandPackagesToBeInstalled $buildRequiredPackageLists_persistent)"
-	fi
+	PACKAGES_PERSISTENT="$buildRequiredPackageLists_persistent"
 	if test -n "$buildRequiredPackageLists_volatile"; then
 		resetIFS
-		for buildRequiredPackageLists_package in $(expandPackagesToBeInstalled $buildRequiredPackageLists_volatile); do
+		for buildRequiredPackageLists_package in $buildRequiredPackageLists_volatile; do
```
mcfedr: I think this whole `for` becomes useless.

mlocati:
What if, in Debian, a package is listed both in the volatile and in the persistent list?

mlocati:
Or, what if somebody calls `apk add` / `apt-get install` before calling this script to add a package that's also in our "volatile" package list? We'll remove it, and that's not fair :wink:

mcfedr:
You see, these are the good questions that I didnt think about :)

- With apk this isnt an issue, as removing the virtual package wont remove deps added before hand
- With debian, yes, more complicated. Need to think some more...
- And yes, if a package was accidently listed in both volatile and persistent it would be removed - I actually hesitated about removing the loop that handled this, but figured it would show in testing that such a mistake had been made.

mlocati:
> With debian, yes, more complicated. Need to think some more...

The current approach is faily simple:
1. determine the list of every APK/APT package required by the extensions we are going to install, separating the packages that are required only at compile time (the "volatile" ones, which can be removed at the end of the script) and the package that are required at runtime (the "persistent" ones)
2. ask APK/APT to list the packages that would be actually installed starting from the "persistent" list (this step will also expand packages required by packages, and omit the packages that are already installed)
3. ask APK/APT to list the packages that would be actually installed starting from the "volatile" list (this step will also expand packages required by packages, and omit the packages that are already installed)
4. remove from the expanded "volatile" list the packages included in the expanded "persistent" list
5. install both the volatile and the persistent packages
6. install the PHP extensions
7. remove the "volatile" packages (as filtered by point 4)

I don't think that there's a safer way to do this that works for both Alpine and Debian (given also that the script may be executed in an environment that can also be already customized by users).

> if a package was accidently listed in both volatile and persistent it would be removed - I actually hesitated about removing the loop that handled this, but figured it would show in testing that such a mistake had been made.

That's not a case covered by tests. This library only have tests for extensions specified in the `Test:` commit message (or triggered by changes in the `data/supported-extensions` file): it's already super-slow, and that's not practical to wait a couple of hours for every test cycle (and it's likely that TravisCI jobs times out).

mcfedr:
Yes, I think your current approch works well, and covers all the cases. `apk` has a really simple way to do this, but unfortunately with `apt` its not that simple...

PR #63, opened 2019-12-18
[Debian 9] libssh-dev conflicts with libssl-dev, use libssl1.0-dev instead

itscaro:
Test: amqp+zip

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -181,9 +181,19 @@ buildRequiredPackageLists () {
 	buildRequiredPackageLists_volatile=''
 	buildRequiredPackageLists_distro="$(getDistro)"
 	buildRequiredPackageLists_phpv=$1
-	if test "$buildRequiredPackageLists_distro" = 'alpine'; then
-		buildRequiredPackageLists_volatile="$PHPIZE_DEPS"
-	fi
+	case "$buildRequiredPackageLists_distro" in
+		alpine)
+			buildRequiredPackageLists_volatile="$PHPIZE_DEPS"
+			;;
+		debian)
+			if test -n "$(apt-cache search libssl1.0 | grep -E ^libssl1\.0)"; then
+				# Debian 9, uses libssl1.0 instead of libssl, due to conflict with libssh-dev
+				buildRequiredPackageLists_libssldev='libssl1.0-dev'
+			else
+				buildRequiredPackageLists_libssldev='libssl-dev'
+			fi
+			;;
+	esac
 	while :; do
 		if test $# -lt 2; then
 			break
@@ -326,7 +336,7 @@ buildRequiredPackageLists () {
 				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libressl-dev cyrus-sasl-dev"
 				;;
 			mongo@debian)
-				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libssl-dev libsasl2-dev"
+				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile $buildRequiredPackageLists_libssldev libsasl2-dev"
 				;;
 			mongodb@alpine)
 				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent icu-libs libsasl snappy"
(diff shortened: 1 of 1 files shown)


install-php-extensions
```diff
@@ -41,6 +41,22 @@ getDistro () {
 	fi
 }
 
+# Get the distribution name with version
```
itscaro: might be handy to do special case for a specific version of OS

mlocati:
When does this conflict occur? A combination of PHP extensions?

install-php-extensions
mlocati: Is this function used somewhere?

itscaro:
I updated yesterday to the new version of this script and for zip extension there is libssl-dev which is newly added.

libssl-dev works fine on buster but not in stretch.

To reproduce the conflict, installation of amqp + zip should be enough.

install-php-extensions
itscaro: Initially I used this function to switch to libssl1.0-dev in stretch but after that I change to use grep like other conditions for alpine. I will remove the function if you think there is no use for it.

mlocati:
> To reproduce the conflict, installation of amqp + zip should be enough.

Everything seems fine for me... Here's a sample session:

```sh
$ docker run --rm -it php:7.3-cli bash

$ curl -o /usr/local/bin/install-php-extensions https://raw.githubusercontent.com/mlocati/docker-php-extension-installer/master/install-php-extensions
$ chmod a+x /usr/local/bin/install-php-extensions
$ install-php-extensions amqp zip
$ php -m | grep -E 'amqp|zip'
amqp
zip
$ php --ri amqp

amqp

Version => 1.9.4
[...]
$ php --ri zip

zip

Zip => enabled
Zip version => 1.15.4
[...]

$ php -r '$zip = new ZipArchive(); var_dump($zip->status);'
int(0)

$ php -r '$connection = new AMQPConnection(); var_dump($connection->isConnected());'
bool(false)
```

itscaro:
```docker run --rm -it php:7.3-cli cat /etc/debian_version
10.2```

This bug is only on stretch - Debian 9, you can try with `docker run --rm -it --entrypoint=bash php:7.3-cli-stretch`

**Comment by mlocati** [comment]
> Test: amqp+zip

This only works in git commit messages, not in the GitHub pull request message

**Comment by mlocati** [comment]
@itscaro don't worry about the issues in this pull request: I'm going to fix them

**Comment by mlocati** [comment]
@itscaro [tests passed](https://travis-ci.org/mlocati/docker-php-extension-installer/builds/627279062?utm_source=github_status&utm_medium=notification), and IMHO this PR is ready to be merged. What do you think?

**Comment by itscaro** [comment]
> @itscaro [tests passed](https://travis-ci.org/mlocati/docker-php-extension-installer/builds/627279062?utm_source=github_status&utm_medium=notification), and IMHO this PR is ready to be merged. What do you think?

@mlocati Thank you for your work, LGTM :)

**Comment by mlocati** [comment]
@itscaro and thanks to you for having detected this issue :wink:

File: install-php-extensions [hunk 2]
```diff
@@ -181,9 +181,19 @@ buildRequiredPackageLists () {
...
-	if test "$buildRequiredPackageLists_distro" = 'alpine'; then
-		buildRequiredPackageLists_volatile="$PHPIZE_DEPS"
-	fi
+	case "$buildRequiredPackageLists_distro" in
+		alpine)
+			buildRequiredPackageLists_volatile="$PHPIZE_DEPS"
+			;;
+		debian)
+			if test -n "$(apt-cache search libssl1.0 | grep -E ^libssl1\.0)"; then
+				# Debian 9, uses libssl1.0 instead of libssl, due to conflict with libssh-dev
```
itscaro: `Debian less than 10` since jessie is included, I forgot about jessie when writing this comment

itscaro: :( Sorry, just tested with jessie, this does not work on jessie

`E: Couldn't find any package by regex 'libssl1.0-dev'`

itscaro:
@mlocati 

this PR breaks imap

on stretch:
```
FAILED TO LIST THE WHOLE PACKAGE LIST FOR
libssl1.0-dev libsasl2-dev libc-client-dev libkrb5-dev 
COMMAND OUTPUT:
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 libc-client2007e-dev : Depends: libssl-dev but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
E: apt-get failed
```

PR #65, opened 2019-12-18
Add support for testing more than one PHP extension at once

mlocati:


mlocati:
Merged as 2de4f3462dab367ec8e2a21affee401bdb9cd27b

PR #74, opened 2019-12-20
fix libssl for alpine

itscaro:


install-php-extensions
```diff
@@ -193,6 +193,11 @@ buildRequiredPackageLists () {
 	case "$buildRequiredPackageLists_distro" in
 		alpine)
 			buildRequiredPackageLists_volatile="$PHPIZE_DEPS"
+			if test -z "$(apk info 2>/dev/null | grep -E ^libssl)"; then
+				buildRequiredPackageLists_libssl="libssl1.0"
+			elif test -z "$(apk info 2>/dev/null | grep -E '^libressl.*-libtls')"; then
+				buildRequiredPackageLists_libssl="libressl2.7-libtls"
+			fi
```
mlocati: If both conditions are not satisfied, `buildRequiredPackageLists_libssl` is undefined, and the script will break (see the `set -o nounset` option at the beginning of the file.

PR #83, opened 2019-12-23
Fix mongo on jessie

mlocati:


mlocati:
I'm going to submit a PR that fixes all the extensions that rely on libssl on Debian.

PR #116, opened 2020-02-15
Don't upgrade Debian packages

mlocati:
This should speed up tests a lot


mlocati:
This change almost halves the test time on Debian Jessie (compare [before](https://github.com/mlocati/docker-php-extension-installer/runs/448100681) and [after](https://github.com/mlocati/docker-php-extension-installer/runs/448107738)).

PR #120, opened 2020-02-17
Add support for ffi

hevertonfreitas:


mlocati:
Why only PHP 7.4? ffi should be available for any php 5.0+ version.
Furthermore: there's no need to update the README.md file: it's updated automatically accordingly to data/supported-extensions

hevertonfreitas:
FFI support was added on PHP 7.4: [https://wiki.php.net/rfc/ffi](https://wiki.php.net/rfc/ffi). I'm not familiar with the PECL version, and the extension has not been updated for almost [10 years](https://github.com/php/pecl-php-ffi). Since there's now an official PHP version, I think that this should be the supported version, right?

mlocati:
> FFI support was added on PHP 7.4: https://wiki.php.net/rfc/ffi. I'm not familiar with the PECL version

I wasn't aware of that. :+1:

PR #146, opened 2020-05-24
Add compatibility memcached with msgpack

fedotov-as:
Fixed `Unable to load dynamic library '/usr/local/lib/php/extensions/no-debug-non-zts-20160303/memcached.so' - Error relocating /usr/local/lib/php/extensions/no-debug-non-zts-20160303/memcached.so: php_msgpack_serialize: symbol not found in Unknown on line 0`

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1304,7 +1304,11 @@ installPECLModule() {
 	case "$2" in
 		http)
 			# http must be loaded after raphf and propro
-			docker-php-ext-enable --ini-name "zzz-$2.ini" "$2"
+			docker-php-ext-enable --ini-name "xx-php-ext-$2.ini" "$2"
+			;;
+		memcached)
+			# memcached must be loaded after msgpack
+			docker-php-ext-enable --ini-name "xx-php-ext-$2.ini" "$2"
 			;;
 		*)
 			docker-php-ext-enable "$2"


mlocati:
Does this error occur when both memcached and msgpack are installed with install-php-extensions?

install-php-extensions
```diff
@@ -1304,7 +1304,11 @@ installPECLModule() {
 	case "$2" in
 		http)
 			# http must be loaded after raphf and propro
-			docker-php-ext-enable --ini-name "zzz-$2.ini" "$2"
+			docker-php-ext-enable --ini-name "xx-php-ext-$2.ini" "$2"
```
mlocati, line 1307: Why did you replace `zzz` with `xx-php-ext`?

fedotov-as:
> Does this error occur when both memcached and msgpack are installed with install-php-extensions?

Right.

Test case on **php:7.1-fpm-alpine**:
`RUN install-php-extensions msgpack memcached`

mlocati:
So, what about amending the commit, adding this line to the commit message?

`Test: msgpack+memcached`

(See https://github.com/mlocati/docker-php-extension-installer/blob/master/README.md#improving-code-for-an-already-supported-extension )

install-php-extensions
fedotov-as: I replaced `zzz` to` xx-php-ext` so that the extensions loading before zz-docker.ini.

If you want, I will rollback this change (for http and memcached).

mlocati: No, that's fine. Thanks for this fix

mlocati:
Thanks!

PR #148, opened 2020-05-24
Update igbinary to 3.1.2

fedotov-as:
Update igbinary to 3.1.2 (latest stable release)


fedotov-as:
@mlocati, please tell me why not used pecl for igbinary? I had no problems.

mlocati:
> please tell me why not used pecl for igbinary?

I absolutely don't remember :wink:
Let's see what happens with #151

mlocati:
Yep, it seems that we don't have any issue by using PECL to install igbinary.

Closing this #148 because I merged #151

PR #155, opened 2020-08-04
add chmod to the odbc configure script

Nemo64:
The build started to fail for me.

I added this chmod statement and now it's working again.


mlocati:
What's your environment? I just submitted a sample pull request that installs the ODBC php extension in every supported docker image, and everything seems to work just fine (see [here](https://github.com/mlocati/docker-php-extension-installer/pull/156/checks)).

Nemo64:
I use it though the [chialab/php](https://github.com/chialab/docker-php) images:

```Dockerfile
FROM chialab/php:7.4-apache

RUN install-php-extensions odbc
```

But I just retried it and now there is no problem so... I don't know, may have been a solar wind or something.

mlocati:
Thanks for the quick feedback

PR #159, opened 2020-08-19
Speedup pecl builds on multicore systems

ederuiter:
Speedup pecl builds by adding MAKE="make -j$(nproc)" so it can make use of more than 1 core

This especially helps speeding up the build of the grpc extension (on my system it reduces the total build time from ~15 minutes down to ~3.5 minutes)

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1339,7 +1339,7 @@ installPECLModule() {
 			printf '  (installing version %s)\n' "$installPECLModule_actual"
 		fi
 		pecl channel-update pecl.php.net || true
-		printf "$installPECLModule_stdin" | pecl install "$installPECLModule_actual"
+		printf "$installPECLModule_stdin" | MAKE="make -j$(nproc)" pecl install "$installPECLModule_actual"
 	fi
 	case "$1" in
 		apcu_bc)


mlocati:
Parallel compilation may not be well supported by all pecl extensions: we'd need to check each one by one...

ederuiter:
Hmm I did not think about that. Would you expect only compile failures or do you expect cases where you can end up with broken extensions without compile failures?  

On the standard php docker images for php:7.4-fpm and php:7.4-fpm-alpine the following extensions are at least compiling without issues:
- amqp 
- bcmath
- bz2
- calendar
- exif
- gd
 gettext
 grpc
 imagick
 intl
 mysqli
 opcache pcntl pdo_mysql protobuf redis soap sockets tidy xdebug xsl yaml zip

mlocati:
So, what about #160 ?

ederuiter:
Works for me :-) 
If you want I can run more compilation checks .. my use case is building our "generic" php images ( https://github.com/thisisdevelopment/php )
Should be quite simple to extend the list of extensions and re-build these images locally

mlocati:
> If you want I can run more compilation checks

Sure! But please add a `Test: <list of touched extensions>` line in the commit messages, so that the GitHub actions know which extensions should be checked (as I did [here](https://github.com/mlocati/docker-php-extension-installer/pull/160/commits) and as described [here](https://github.com/mlocati/docker-php-extension-installer/blob/master/README.md#improving-code-for-an-already-supported-extension)).

BTW let's wait until #160 completes its tests: I'm going to merge it as soon as they complete.

mlocati:
Done, and thanks for the idea!

PR #167, opened 2020-09-18
Support for using pickle as drop in for pecl on php 8

mcfedr:
Still working on testing this, but the idea...

Since 7.4 pecl has been deprecated and the php docker image will no longer enable it from php 8 - https://github.com/docker-library/php/issues/846

[Pickle](https://github.com/FriendsOfPHP/pickle) is meant to be a modern alternative to pecl, so lets give that a go.

A number of extensions, at least sqlsrv and xdebug, cannot be installed with the current version of pickle, looks like this was fixed in https://github.com/FriendsOfPHP/pickle/pull/185


README.md
```diff
@@ -48,89 +48,89 @@ RUN install-php-extensions gd xdebug
 <!-- #  EDIT THE data/supported-extensions FILE INSTEAD        # -->
 <!-- #                                                         # -->
 <!-- ########################################################### -->
-| Extension | PHP 5.5 | PHP 5.6 | PHP 7.0 | PHP 7.1 | PHP 7.2 | PHP 7.3 | PHP 7.4 |
```
mlocati, line 51: You don't need to manually update this README section: it will be updated automatically once the pull request will be merged

install-php-extensions
```diff
@@ -1377,6 +1376,23 @@ installPECLModule() {
 	esac
 }
 
+installPickle() {
+	if ! test "$PHP_MAJMIN_VERSION" = "800"; then
```
mlocati, line 1380: What about moving this PECL related stuff below the `if ! stringInList "$PHP_MODULE_TO_INSTALL" "$BUNDLED_MODULES"; then` line? That way, we'd run `pecl channel-update pecl.php.net || true` only if we'll use PECL.

Also, what about replacing `if ! test "$PHP_MAJMIN_VERSION" = "800"; then` with `if test $PHP_MAJMIN_VERSION -lt 800; then` ?

install-php-extensions
```diff
@@ -1479,12 +1496,14 @@ fi
 
 sortModulesToInstall
 
+docker-php-source extract
```
mlocati, line 1499: Why did you move these line up?

README.md
mcfedr: aha, ok, i just ran the script.

install-php-extensions
mcfedr: Because i use `$BUNDLED_MODULES` in `installPickle` to check if pickle will be required - its only installed when at least one of `$PHP_MODULES_TO_INSTALL` is not a bundled module.

install-php-extensions
mcfedr: Yes, makes sense to move the whole check for php 8 inside the loop.

install-php-extensions
mlocati: Yep, sure, `installPickle` must be after `docker-php-source extract` and `BUNDLED_MODULES=...`,
I was asking if there's a reason to move those 2 (or 3 considering `installPickle`) lines up

mcfedr:
> `$PHP_MAJMIN_VERSION` contains an integer: no needs to use double quotes here...

I've removed it, my habit is to always quote everything when writing shell code, generally its a [safe approach](https://github.com/koalaman/shellcheck/wiki/SC2086), but ill stick to the style

install-php-extensions
mcfedr: Yes- because pickle requires the `zip` module ( the line `PHP_MODULES_TO_INSTALL="zip $PHP_MODULES_TO_INSTALL"` in `installPickle`) which requires `apt`/`apk` packages to be installed - so this section needs to happen before the platform packages are calculated and installed.

mcfedr: Which actually, I'm looking - ill change this to check if zip is already installed or in the list.

install-php-extensions
```diff
@@ -854,6 +861,12 @@ compareVersions() {
 installBundledModule() {
 	printf '### INSTALLING BUNDLED MODULE %s ###\n' "$1"
 	case "$1" in
+		enchant)
+			if test "$DISTRO" = "alpine" && test $DISTRO_VERSION_INT -ge 312 && test $PHP_MAJMIN_VERSION -lt 800; then
+				echo "Cannot install enchant on $DISTRO_VERSION with php $PHP_MAJMIN_VERSION - https://bugs.php.net/bug.php?id=76962"
```
mcfedr, line 866: This is not very convincing, maybe there is a better idea @mlocati ? Basically on alpine 3.12 there is only enchant2, but php<8 doesnt support enchant2, so on this version of alpine its impossible to enable enchant

install-php-extensions
```diff
@@ -1377,6 +1383,27 @@ installPECLModule() {
...
+		if ! stringInList "$PHP_MODULE_TO_INSTALL" "$BUNDLED_MODULES"; then
+			if test $PHP_MAJMIN_VERSION -lt 800; then
+				pecl channel-update pecl.php.net || true
+				PECL_COMMAND="pecl install"
+				return
+			fi
+			curl -L -f https://github.com/FriendsOfPHP/pickle/releases/latest/download/pickle.phar -o /tmp/pickle
+			chmod +x /tmp/pickle
+			if ! stringInList 'zip' "$(getPHPInstalledModules)"; then
+				PHP_MODULES_TO_INSTALL="zip $(removeStringFromList 'zip' "$PHP_MODULES_TO_INSTALL")"
```
mlocati, line 1399: Do we need to remove `zip` if it's not included?

mlocati: Sorry, your code is correct.

mcfedr: if your asking - i wondered about trying to remove zip in clean up if the user didnt ask for it , but i think that would be quite complicated to add to the code as there is no support for removing extensions at the moment.

mlocati:
It seems that the `jessie` tests don't want to start.
Since nothing here should affect `jessie` stuff, I'd merge this PR anyway... thoughts?

mcfedr:
> It seems that the `jessie` tests don't want to start.
> Since nothing here should affect `jessie` stuff, I'd merge this PR anyway... thoughts?

I'm pretty happy with the result, I think it would be great to merge it!

mlocati:
Yep, you did a great work here @mcfedr !
It's a bit late in my time zone now (CEST).
Tomorrow morning I'll merge this great PR even if jessie tests won't be executed.

PR #173, opened 2020-10-14
Install the beta version of xdebug 3

mlocati:
See https://xdebug.org/announcements/2020-10-14


mlocati:
See https://bugs.xdebug.org/view.php?id=1863

PR #182, opened 2020-11-05
Update README.md

4n70w4:
`chmod +x` gets the same result that `chmod a+x` and `chmod uga+x`.

> Classes of users are used to distinguish to whom the permissions apply. If no classes are specified "all" is implied. The classes are represented by one or more of the following letters:

Reference | Class | Description
------------ | ------------- | -------------
u | user | file owner
g | group | members of the file's group
o | others | users who are neither the file's owner nor members of the file's group
a | all | all three of the above, same as ugo


mlocati:
Thanks!

Issue #130, opened 2020-03-13
install specific version


michabbb:
hi, is there anything I missed, that gives me the chance to install a specific version of an extension ???

thanks

mlocati:
Nope, it's not (yet) supported

michabbb:
🆗 thanks

glen-84:
Can this be left open as a feature request?

piotrkochan:
How about this year?

mlocati:
I'm planning to implement this, as soon as I have some spare time

nesk:
If it can help, I install a specific version of `rdkafka` by monkey-patching the script:

```
COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/bin/
RUN sed -i -e $'s/rdkafka)/rdkafka)\\\n\\\t\\\t\\\tinstallPECLModule_actual="rdkafka-3.1.3"\\\n\\\t\\\t\\\t;;\\\n\\\t\\\trdkafka-latest)/' /usr/bin/install-php-extensions
RUN install-php-extensions rdkafka
```

It does the following replacements:

```diff
		raphf)
			if test $PHP_MAJMIN_VERSION -lt 700; then
				installPECLModule_actual="$1-1.1.2"
			fi
			;;
		rdkafka)
+			installPECLModule_actual="rdkafka-3.1.3"
+			;;
+		rdkafka-latest)
			if test $PHP_MAJMIN_VERSION -le 505; then
				installPECLModule_actual="$1-3.0.5"
			else
				installPECLModule_tmp=
				case "$DISTRO" in
```

It's crappy, but it works until the version feature is available.

cytodev:
@mlocati This could possibly be resolved by changing [this line](https://github.com/mlocati/docker-php-extension-installer/blob/4157d994c964e6bece1a5413998091d0ce28a6a8/install-php-extensions#L1085) to split on hypens. Versions are usually notated behind the package name, separated by a hyphen. That way you can supply the installer versions to install (which pecl accepts) and strip them to enable them with `docker-php-ext-enable`.

mrsrvman:
https://github.com/symfony/symfony/issues/39195

evs-xsarus:
For XDEBUG, I've solved this temporary with:

```
# Explicitly install Xdebug 2.x as 3.x is now installed by mlocati's script. That version breaks Magento.
# Als install it after mlocati's script, that script will install all required packages for pecl and make.
[...]

PR #188, opened 2020-11-26
Strip everything after - char to allow installation of specific versions

cytodev:
Versions are usually notated behind the package name, separated by a hyphen. That way you can supply the installer versions to install (which pecl accepts) and strip them to enable them with docker-php-ext-enable.

See issue #130

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1082,7 +1082,7 @@ installModuleFromSource() {
 	./configure $3 CFLAGS="${4:-}"
 	make -j$(getProcessorCount) install
 	cd - >/dev/null
-	docker-php-ext-enable "$1"
+	docker-php-ext-enable "${1%-*}"
 	case "$1" in
 		snuffleupagus)
 			cp -a "$installModuleFromSource_dir/config/default.rules" "$PHP_INI_DIR/conf.d/snuffleupagus.rules"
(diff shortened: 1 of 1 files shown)


mlocati:
It's not that easy, since we have a lot of lines what check the package names:

- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L59-L64) we normalize the package names
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L134) we check for duplicated package names provided by the users
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L136) we check for already installed packages
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L180-L198) we check for package dependencies
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L239) we need to strip out the package versions
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L824) we need to strip out the package versions
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L910) we need to strip out the package versions
- [here](https://github.com/mlocati/docker-php-extension-installer/blob/1.0.3/install-php-extensions#L1103) we already may set the package version: we need to let users override it

Furthermore, pickle is having a lot of problems while dealing with versions: we need to wait until they fix it before doing anything with versions (see https://github.com/FriendsOfPHP/pickle/issues/188 and https://github.com/FriendsOfPHP/pickle/issues/191)

mlocati:
Furthermore, when submitting a pull request, please specify some tests to be performed (see [here](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension)).

cytodev:
I agree that this is indeed a very quick patch to define version constraints.

Based on my specific use-case (wanting to install xdebug-2.9.8) this covers all parts required for the script to install the specified version and enable the normalized shared object.

I had not looked into the specifics and just quickly patched it on my end; sharing it in case it could be helpful. You could probably pull the same trick I did with the string substitution to isolate specific version numbers from their package names and pass those two around where needed.

I'll let you do the honours of closing this PR if it's not helping anything ;)

Have a great day!

mlocati:
I'd actually adopt a different approach: instead of changing all the places where we use package names, I'd do that
- for every argument received, extract the provided version (if any) and save it somewhere somehow
- keep most of the code like it's now
- when we actually install a package, check if the users provided a version for it

That would simplify the changes a lot. But I'd wait until they fix the pickle issues...

mlocati:
I'm working on an alternative approach right now.

PR #190, opened 2020-11-28
Add support for Mosquitto with PHP 8.0

mlocati:


mlocati:
Nope, mosquitto doesn't support PHP 8.0 yet.

PR #224, opened 2020-12-15
Install OpenSSL 1.1.1i instead of the system default one

mlocati:


mlocati:
This OpenSSL stuff is a real mess, it doesn't seem it's possible to side-install OpenSSL 1.1.1i alongside the system default one because of system packages dependencies...

PR #268, opened 2021-01-21
Fix installing smbclient on old Alpine Linux versions

mlocati:


mlocati:
See also https://github.com/eduardok/libsmbclient-php/issues/79

Issue #271, opened 2021-01-24
Add PHP 8.1 support


mvorisek:
PHP 8.1 alpha2 is available, feature request to add PHP 8.1 support.

mlocati:
Adding support for a new php version requires tons of hours. And working with a development version would require to do that whenever something changes.
So, I'm not going to officially add support for 8.1-dev, but I'd of course accept pull requests that fix support for 8.1 (provided that those changes don't break support for the current php versions).

PR #272, opened 2021-01-24
Add support/testing for PHP 8.1

mvorisek:
fixes https://github.com/mlocati/docker-php-extension-installer/issues/271


mvorisek:
I wonder, how can I add testing against `phpdaily/php` docker images?

mlocati:
I guess tests won't complete because of timeout

mvorisek:
> 
> 
> I guess tests won't complete because of timeout

related with: https://github.com/mlocati/docker-php-extension-installer/issues/275, to overcome the timeout I now added support for extensions provided directly from php sources, which should always compile. Once this PR is merged, I will add support for other extensions in another PR.

mvorisek:
given the efford to maintain this and not having full support to mark specific job as informative only - see https://github.com/actions/toolkit/issues/399 , I am closing this for now

mlocati:
Yep, I'd wait that PHP 8.1 becomes a bit more stable (that is, at least in the beta stage).

mvorisek:
@mlocati PR is passing, most of the extensions are from php-src, thus the build should always pass, can we merge it to reduce the testing time of other extensions?

mlocati:
Sure! I'll wait that those last tests pass

mvorisek:
The CI should be passing, `Check on alpine3.12` issue seems to be related with cached PHP 8.1 alpha1 image, alpha2 is fixed.

mlocati:
> The CI should be passing, `Check on alpine3.12` issue seems to be related with cached PHP 8.1 alpha1 image, alpha2 is fixed.

I restarted the CI jobs

mvorisek:
Alpine v3.12 images are no longer updated.

See https://github.com/docker-library/php/pull/1166 and related pulled tag https://hub.docker.com/_/php?tab=tags&page=1&ordering=last_updated&name=8.1-rc-cli-alpine3.12

Can you please adjust the test script that PHP 8.1+ is not tested againt Alpine 3.12?

PR #273, opened 2021-01-24
Run CI on push as well

mvorisek:
allow easier development without the need to open PR first

### DIFF
--- a/.github/workflows/test-extensions.yml
+++ b/.github/workflows/test-extensions.yml
@@ -2,6 +2,7 @@ name: Test extensions
 
 on:
   - pull_request
+  - push
 
 jobs:
   check_syntax_data:


mlocati:
Every contributor to this project have to submit pull requests to change files in this project.
Even I always do that (except when I touch irrelevant files like the readme).

So, tests would be executed twice: when people submit a pull request, and when the pull requests get merged. And that's a waste of time.

README.md
```diff
@@ -8,7 +8,7 @@ This repository contains a script that can be used to easily install a PHP exten
 
 The script will install all the required APT/APK packages; at the end of the script execution, the no-more needed packages will be removed so that the image will be much smaller.
 
-Supported docker images are all the Alpine/Debian versions, except for PHP 5.5 where we only support Debian 8 (jessie) (that is, `php:5.5`, `php:5.5-apache`, `php:5.5-cli`, `php:5.5-fpm`, `php:5.5-zts`).
+Supported docker images are all the Alpine/Debian versions, except for PHP 5.5 where we only support Debian 8 (jessie).
```
mlocati, line 11: Any reason to change this line?

mvorisek: I think it is not needed... I would even drop support for PHP 5.5 ;-)

mlocati: No, I won't drop support for PHP 5.5. And I don't see any problem being more explicit about supported versions (it may also save me some time because I won't have to answer questions about exactly supported versions)

mvorisek: I removed this change...

mvorisek:
There is some issue when building on push:

image

mlocati:
As I wrote above, I won't merge this PR, since I don't want tests to ve executed twice

mvorisek:
ok

mvorisek:
ok, but the build issue should be still fixed. a lot of people test changes with `push`, I belive PR into forks also does not work

mlocati:
>  a lot of people test changes with `push`

Are you talking in general, or specifically for this project?
When we run tests on push + pull requests, here's what would happen:

1. I push a new branch to the repo, with the commits I'd like to merge
2. Github sees a push, so it runs the tests
3. I create the pull request by using that branch
4. Github sees a pull request, so it runs the tests
5. I merge the pull request
6. Github sees a push, so it runs the tests

So, the same tests would be executed 3 times.
And for this project, tests may require a lot of time.

mvorisek:
For this project. I respect your decision to not merge this PR, but the build should work in general for forks/push events.

mlocati:
The way the tests are currently implemented requires that we know the list of relevant commits (for example, we automatically tests an extension if its line in the `data/supported-extensions` file changes, or if a commit message contains the line `Test: extension_name`).

But detecting the relevant commits in push events is not so easy:

- if we use `master` as the base ref, we'd require that people dont use `master` to create a pull request, and we'd require that the `master` of the fork is in sync with this source repo
- if we'd only consider the commits included in the push event, we risk to not consider some commits (think for example when people push a second commit after having pushed a first commit).

mlocati:
(And of course you can always create a pull request directly in your own fork)

mvorisek:
Yes, I already did that, that is working: https://github.com/mvorisek/docker-php-extension-installer/pull/1

PR #274, opened 2021-01-25
Add testing with phpdaily-alpine images

mvorisek:
for https://github.com/mlocati/docker-php-extension-installer/pull/272, this should not break anything, as daily builds of bugfix branches should have unchanged ABI

### DIFF
--- a/.github/workflows/test-extensions.yml
+++ b/.github/workflows/test-extensions.yml
@@ -74,6 +74,7 @@ jobs:
           - jessie
           - stretch
           - buster
+          - phpdaily-alpine
     name: Check on ${{ matrix.distro }}
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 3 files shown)


mlocati:
I don't know... is it wise to use nightly builds? Don't we risk to keep fixing stuff because we're running against nightly builds?

Why should we do that?

mvorisek:
For bugfix branches this should be safe, because of:
- ABI freeze
- Docker images should built only when the base sources are tested

We should do this to tests possible issues upfront and to have testing for new php versions earlier than official images are releases.

I think it is safe to merge, if there will be any issues, we can revert.

mlocati:
What's `ABI`?

mvorisek:
https://en.wikipedia.org/wiki/Application_binary_interface - stable C headers (`*.h` files)

mlocati:
So, if nothing changes on the system side, why should we test installing PHP extensions? I'd only test what's changing (and in our case are the new version of PHP extensions - and we already do that)

mvorisek:
Functional tests are never enough :)

mlocati:
Even my spare time 😉 

I don't know, I really don't have time to take care of nightly stuff...

mvorisek:
Well, I personally do not need to pursue this PR now.

PR #291, opened 2021-02-05
Fix typo in script version warning message

yannoff:
Not a huge typo, anyway... :wink:

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -28,7 +28,7 @@ if test "$IPE_VERSION" = master && test "${CI:-}" != true; then
 #                                                                                                           #
 # You are using an unsupported method to get install-php-extensions!                                        #
 #                                                                                                           #
-# Please update the way you fetch it. Read the instrictions at                                              #
+# Please update the way you fetch it. Read the instructions at                                              #
 # https://github.com/mlocati/docker-php-extension-installer#usage                                           #
 #                                                                                                           #
 # For example, if you get this script by fetching                                                           #


mlocati:
Thanks!

yannoff:
> Thanks!

You're welcome :smile_cat: 

By the way, just for the sake of my own acknowledgement... I can't figure out how this part works:

Given the assignment in [line 20](https://github.com/mlocati/docker-php-extension-installer/blob/master/install-php-extensions#L20), it seems to me that the first condition on [line 22](https://github.com/mlocati/docker-php-extension-installer/blob/master/install-php-extensions#L22) always evaluate to `true` ... Am I missing something here ?

mlocati:
The `IPE_VERSION` variable is updated by https://github.com/mlocati/docker-php-extension-installer/blob/1.2.5/.github/workflows/readme-release.yml#L153 (so you should use the file attached to GitHub releases or the Docker image)

mlocati:
You can see how it works by checking this action for instance: https://github.com/mlocati/docker-php-extension-installer/actions/runs/541214743

yannoff:
> You can see how it works by checking this action for instance: https://github.com/mlocati/docker-php-extension-installer/actions/runs/541214743

Ahhh I see :smile_cat:  Thanks for the clarification  :ok_hand:

I realize I reproduced manually the exact anti-pattern that the warning message is here to prevent... :stuck_out_tongue_closed_eyes:

Issue #302, opened 2021-02-17
Cannot install redis after protobuf (php:8.0-alpine) – "Could not determine info from the PHP binary"


caugner:
## Summary

When installing the `redis` extension after installing the `protobuf` extension, the installation fails with the following error:

```
In PHP.php line 275:

  Could not determine info from the PHP binary


install [--no-convert] [--defaults] [--source] [--with-configure-options WITH-CONFIGURE-OPTIONS] [--save-logs SAVE-LOGS] [--dry-run] [--php PHP] [--ini INI] [--tmp-dir TMP-DIR] [--version-override [VERSION-OVERRIDE]] [--] [<path>]
```

## Dockerfile

```Dockerfile
FROM php:8.0-cli-alpine

RUN apk add --no-cache git && \
    curl -sSL -o /usr/local/bin/install-php-extensions https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions && \
    chmod uga+x /usr/local/bin/install-php-extensions && \
    sync && \
    install-php-extensions protobuf redis && \
    rm /usr/local/bin/install-php-extensions
```

## Build log

```
Sending build context to Docker daemon  420.4kB

Step 1/2 : FROM php:8.0-cli-alpine
 ---> 9484fd084dd9
Step 2/2 : RUN apk add --no-cache git &&     curl -sSL -o /usr/local/bin/install-php-extensions https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions &&     chmod uga+x /usr/local/bin/install-php-extensions &&     sync &&     install-php-extensions protobuf redis &&     rm /usr/local/bin/install-php-extensions
 ---> Running in a68adde10ba8
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/community/x86_64/APKINDEX.tar.gz
(1/3) Installing expat (2.2.10-r1)
(2/3) Installing pcre2 (10.36-r0)
(3/3) Installing git (2.30.1-r0)
Executing busybox-1.32.1-r2.trigger
OK: 24 MiB in 35 packages
install-php-extensions v.1.2.13
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/main/x86_64/APKINDEX.tar.gz
[...]

PR #303, opened 2021-02-17
Let's test pickle when PHP shows warnings

mlocati:


mlocati:
The new pickle version (0.7.1) fixes #302

PR #324, opened 2021-03-21
Readme: Remove "on the fly example"

glensc:
The `COPY` solution is what should be used because it allows the layer to be cached.

The other solution included a weird `sync` hack, which without an actual bug to investigate should be removed:
- https://github.com/mlocati/docker-php-extension-installer/commit/eca94d432b66589fed91b79c1b8f5bdcfca45d26


mlocati:
Many people use that approach. If you prefer the COPY method simply use it

PR #330, opened 2021-04-23
Check installer signature when installing composer

### LINKED ISSUE #329: Interest check: validate composer SHA
As per the official docs, the programmatic way to install composer is with the following script:

```sh
#!/bin/sh

EXPECTED_CHECKSUM="$(php -r 'copy("https://composer.github.io/installer.sig", "php://stdout");')"
php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
ACTUAL_CHECKSUM="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"

if [ "$EXPECTED_CHECKSUM" != "$ACTUAL_CHECKSUM" ]
then
    >&2 echo 'ERROR: Invalid installer checksum'
    rm composer-setup.php
    exit 1
fi

php composer-setup.php --quiet
RESULT=$?
rm composer-setup.php
exit $RESULT
```

Source: https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md#how-do-i-install-composer-programmatically-

Would you be open to a PR that implements the SHA check? This spreads the security out so that both composer and github would have to be compromised for someone to inject malicious code into the downloaded composer binary (and we've see how this is possible recently with the recent PHP git compromise).

Happy to give it a whirl myself, but wanted to check if the maintainers would be open to a PR before I spend the time on it.

Thanks for the fantastic project.

mlocati:
Closes #329

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1289,7 +1289,35 @@ installComposer() {
 		fi
 	fi
 	printf '### INSTALLING %s ###\n' "$installComposer_fullname"
-	curl -sSLf https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer $installComposer_flags
+	actuallyInstallComposer /usr/local/bin composer "$installComposer_flags"
+}
+
+# Actually install composer
+#
+# Arguments:
+#   $1: the directory where composer should be installed (required)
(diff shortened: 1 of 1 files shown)


timacdonald:
This looks great! Thanks so much for making this update. Big +1 from me.

PR #363, opened 2021-06-30
Use maven instead of ant when building zookeeper

pataar:
Ant is not used anymore: https://stackoverflow.com/questions/64916994/build-xml-does-not-exist-when-install-zookeeper-client-c#comment116094374_65657279

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -2199,16 +2199,32 @@ installRemoteModule() {
 			fi
 			;;
 		zookeeper)
+			if test -z "$installRemoteModule_version"; then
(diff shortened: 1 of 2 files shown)


install-php-extensions
```diff
@@ -48,7 +48,7 @@ fi
 
 # Reset the Internal Field Separator
 resetIFS() {
-	IFS='
```
mlocati, line 51: This must NOT be changed

pataar: Reverted!

mlocati:
Thanks!

PR #386, opened 2021-07-10
Feature - add new extension xmldiff

llaville:


install-php-extensions
```diff
@@ -2238,6 +2244,10 @@ installRemoteModule() {
 				fi
 			fi
 			;;
+		xmldiff)
+			# whether to use system diffmark library
+			addConfigureOption 'with-libdiffmark' 'yes'
```
mlocati, line 2249: I don't see any `with-libdiffmark` option: see https://mlocati.github.io/pecl-info/#xmldiff

install-php-extensions
```diff
@@ -926,6 +926,12 @@ buildRequiredPackageLists() {
 			xlswriter@debian)
 				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile zlib1g-dev"
 				;;
+			xmldiff@alpine)
+				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libxml2-dev"
```
mlocati, line 930: It seems we also need `libstdc++ ` (persistent)

install-php-extensions
llaville: See it on source code http://svn.php.net/viewvc/pecl/xmldiff/tags/release_1.1.3/config.m4?revision=352303&view=markup
Perharps I misunderstood

install-php-extensions
llaville: probably : i'm sorry but I dind't run the script locally and test it first (it's very strange behavior)

mlocati: That's why we have that GitHub Action 😉

install-php-extensions
mlocati: For such options we have to look into the package.xml file (or use that pecl-info website)

install-php-extensions
llaville: Finally found it how to test it locally.

install-php-extensions
llaville: Don't worry about it, because I found that `libdiffmark` is not supported with PHP 8.x (so I've disabled it)

mlocati:
@llaville PECL won't be able to set the `with-libdiffmark` configure option, since it's not defined in the package.xml file.
Furthermore, this option is not meant to enable/disable the usage diffmark: its value simply tells the compiler where the libdiffmark library can be found. If we don't specify this value (or if we specify `no`), the compiler will use the bundled libdiffmark.
Indeed, if we compile the extension without specifying this flag, we have:

```
/ # php --ri xmldiff

xmldiff

xmldiff support => enabled
xmldiff version => 1.1.3
diffmark library => 0.10 bundled
Revision => $Id: xmldiff.cpp 350542 2020-09-10 12:08:03Z remi $
```

PR #392, opened 2021-07-17
Feature - Add support for yac

llaville:
[YAC](https://github.com/laruence/yac) is an alternative to APC and memcached


mlocati:
I'm sorry I'm a couple of thousands Km away from my PC, I'll review this PR when I'll be back

PR #399, opened 2021-07-30
Fix uopz for < php 8

mcfedr:
A new release of uopz only supports php >8, so now installs are failing for older versions


mlocati:
Thanks!

PR #414, opened 2021-08-07
Add SPX support

1ed:
Add support for SPX https://github.com/NoiseByNorthwest/php-spx


1ed:
Is it OK to use a commit from master instead of a tag? The issue was fixed in https://github.com/NoiseByNorthwest/php-spx/issues/121, but there was no new release since than. (https://github.com/NoiseByNorthwest/php-spx/issues/174)

mlocati:
Sure! We can do something like what we had in https://github.com/mlocati/docker-php-extension-installer/pull/232/files

mlocati:
That's perfect!

1ed:
Thanks!

PR #421, opened 2021-08-18
Install the correct version of blackfire on alpine

dsech:
The regular "linux" build for blackfire extension is incompatible with alpine linux. Attempting to start a profile will cause a segfault in the php(-fpm) process.

But blackfire comes with dedicated builds for alpine linux, so we should use these.

Also, I suppose a test that catches this issue would be helpful. But it requires to actually trigger a profile request in order to cause the segfault, which sounds complicated to do.

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1666,10 +1666,18 @@ installRemoteModule() {
 					installRemoteModule_tmp1=amd64
 					;;
 			esac
+			case $DISTRO in
+				alpine)
+					installRemoteModule_distro=alpine
+					;;
+				*)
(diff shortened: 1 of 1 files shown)


mlocati:
@dsech Could you amend your last commit, by specifying the following commit message?

```
Install the correct version of blackfire on alpine

Test: blackfire
```

That way, [as specified in the README file](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension), we tell the CI system to test the blackfire extension.

dsech:
Hi @mlocati , I've ammended the commit message

PR #445, opened 2021-09-29
Support uploadprogress for PHP 8.0

llaville:
Tested with PHP 7.4.24 and 8.0.11


mlocati:
Thanks!

mlocati:
It seems that uploadprogress is also compatible with PHP 8.1rc. I'll submit another PR to check that

PR #475, opened 2021-12-06
Add libldap-common to debian dependencies

### LINKED ISSUE #474: Secure LDAP doesn't work on debian 11+
The libldap-common package isn't installed when adding the php ldap plugin on Debian Bullseye.  This means there isn't an ldap.conf installed, which means ldap in php isn't able to read the systemwide CA database, meaning secure LDAP (the only way LDAP should be used in 2021!) doesn't work.  

In debian <10 this was installed implicitly as part of another package: https://github.com/docker-library/php/issues/1194#issuecomment-904954896

I have reported this as a bug to [docker-php](https://github.com/docker-library/php/issues/1228), and they suggested it should be fixed here.

My team have encountered this issue in at least two pieces of PHP software since the softwares have moved their base containers to debian 11: [roundcube](https://github.com/roundcube/roundcubemail-docker/issues/160) [nextcloud](https://github.com/nextcloud/docker/issues/1572), oo it'd be good to get it sorted in the extension installer!

thinkl33t:
This used to be installed implicitly within debian-based docker containers by curl, but as of debian 11 isn't.

This is needed to let secure LDAP work out of the box.

Fixes #474

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -710,6 +710,7 @@ buildRequiredPackageLists() {
 				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile openldap-dev"
 				;;
 			ldap@debian)
+				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent libldap-common"
 				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libldap2-dev"
 				;;
 			maxminddb@alpine)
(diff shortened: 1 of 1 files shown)


mlocati:
Ad [described in README.md](https://github.com/mlocati/docker-php-extension-installer#changing-the-supported-php-versions-for-an-already-supported-php-extension), could you add a line like the following line to the commit message?

```
Test: ldap
```

thinkl33t:
> Ad [described in README.md](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension), could you add a line like the following line to the commit message?
> 
> ```
> Test: ldap
> ```

Done.

mlocati:
libldab-common doesn't seem to exist on Debian Jessie: I've updated this PR (see 779525c437540a9e298899af41384c412c9fc381)

thinkl33t:
Ah excellent thank you, i was just doing that myself :)

PR #559, opened 2022-04-02
Support tensor on PHP 8.1

mlocati:


mlocati:
See https://github.com/RubixML/Tensor/issues/28

PR #590, opened 2022-06-11
Add sockets PHP 8.2 support

### LINKED ISSUE #582: php-src master does not build on Alpine
### Version of install-php-extensions

latest

### Error description

see https://github.com/php/php-src/issues/8681

the `sockets` ext needs newly `linux-headers` package on Alpine during the install

### Docker image

latest

### Minimal Dockerfile

_No response_

mvorisek:
fixes #582


mlocati:
It's needed only fot php 8.2+, isn't it?

mvorisek:
> It's needed only fot php 8.2+, isn't it?

Yes, needed only for PHP 8.2+. I tried to add a `if test $PHP_MAJMIN_VERSION -ge 820; then` cond, but it seems it does not work, do you know how to fix it?

mlocati:
> `if test $PHP_MAJMIN_VERSION -ge 820; then`

Try with 802 instead of 820

mlocati:
See https://github.com/mlocati/docker-php-extension-installer/blob/1.5.21/install-php-extensions#L82

mvorisek:
> > `if test $PHP_MAJMIN_VERSION -ge 820; then`
> 
> Try with 802 instead of 820

👍, thanks for the support

PR #597, opened 2022-06-16
Add support for snappy

mcfedr:


install-php-extensions
```diff
@@ -1097,6 +1097,14 @@ buildRequiredPackageLists() {
 				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent libsmbclient"
 				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libsmbclient-dev"
 				;;
+			snappy@alpine)
+				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent snappy"
+				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile snappy-dev"
+				;;
+			snappy@debian)
+				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent libsnappy1v5"
```
mlocati, line 1105: Try with
```sh
buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent ^libsnappy1(v[0-9]+)?$"
```

data/supported-extensions
```diff
@@ -85,6 +85,7 @@ redis 5.5 5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2
 seaslog 5.5 5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2
 shmop 5.5 5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2
 smbclient 5.5 5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2
+snappy 5.6 7.0 7.1 7.2 7.3 7.4 8.0 8.1 8.2
```
mlocati, line 88: It seems that we can even use snappy on PHP 5.5

mcfedr: Making doubt not putting it in lz4 https://github.com/mlocati/docker-php-extension-installer/pull/596/files as well, i thought i checked this locally, but clearly not very well

mlocati: Don't worry, I'll fix it :wink:

PR #598, opened 2022-06-16
Support lz4 on PHP 5.5 too

mlocati:


mcfedr:
I had two issues with this

on debian (`php:5.5`)

compiles but doesnt run:

```
php > lz4_compress("a");
php: symbol lookup error: /usr/local/lib/php/extensions/no-debug-non-zts-20121212/lz4.so: undefined symbol: LZ4_compress_default
root@a49cee0005a9:/#
```

and on alpine `php:5.5-alpine` I had to install `lz4` instead of `lz4-libs`

mcfedr:
like this for alpine worked, https://github.com/mlocati/docker-php-extension-installer/compare/master...mcfedr:lz455?expand=1

mcfedr:
yea, I tried with the older version of lz4-ext and it didnt work either - I dont think its the end of the world to not support 5.5 - I only tried 5.6 because i was interested...

PR #603, opened 2022-06-21
Fix parallel compilation with pickle

ederuiter:
The MAKE environment variable seems to be only supported on pecl, not with pickle
Use the MAKEFLAGS environment variable which is automatically passed to the child process and is directly supported by make


mlocati:
Nice catch!

PR #605, opened 2022-06-24
Fix intl on Alpine 3.16 by installing icu-data-full

### LINKED ISSUE #604: Intl provides only "en" support on alpine 3.16
Since aline 3.16, the ICU package is split into 2 sub-set (https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.16.0):
- icu-data-en 
- icu-data-full

By default, the intl extension requires `icu-libs` which requires `icu-data-en`.
As a result, the intl extension provides support only for `en` locale.

I think the fix should be something like that, but I'm not confident enough in the script syntax to submit a PR..

```
if test $DISTRO_MAJMIN_VERSION -ge 316; then
	buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent icu-data-full icu-libs"
else
	buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent icu-libs"
fi
```

mlocati:
Close #604

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -608,6 +608,10 @@ buildRequiredPackageLists() {
 			else
 				buildRequiredPackageLists_libssldev='libretls-dev'
 			fi
+			buildRequiredPackageLists_icuPersistent=''
+			if test $DISTRO_MAJMIN_VERSION -ge 316; then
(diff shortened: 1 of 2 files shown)


jderusse:
thank you

PR #612, opened 2022-07-15
Let users install a smaller but English-only ICU lib

### LINKED ISSUE #611: Allow to instal `icu-data-en` only
this is a follow up future request to https://github.com/mlocati/docker-php-extension-installer/pull/605

https://pkgs.alpinelinux.org/package/edge/main/x86/icu-data-full requires 30 MB which equals to about 10% of the total Alpine image size

please add a switch to allow to instal `icu-data-en` only

mlocati:
Close #611


mvorisek:
thank you ❤️

PR #620, opened 2022-08-03
Add support for swoole on PHP 8.2

mlocati:


mlocati:
See https://github.com/swoole/swoole-src/issues/4781

PR #658, opened 2022-11-17
Update README.md

Legion112:
A better way to run this beautiful script.


mlocati:
Yep, nice approach. Could you use the [suggestion](https://github.com/mlocati/docker-php-extension-installer/pull/658#discussion_r1025290178) by @jdreesen?

PR #676, opened 2022-12-21
Use R2 URL for ext-relay

tillkruss:
Hey 👋

We're moving from S3 to R2 in the next couple of months. The old S3 links will keep working, but all new version will be published to R2.


mlocati:
Thanks!

PS: I've added the empty commit 7afbafc3cd51f8a45618d7f950a1bbbe07f8e23d to tell the GitHub Actions to test installing the extension, as described [here](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension)

PR #677, opened 2022-12-27
Added direct execution from curl

GreyXor:
I added a part in the readme that indicates how to use the script without having to store it, chmod it and delete it


GreyXor:
Hello, no it's for alpine, but I can change it to sh, it will work the same

PR #709, opened 2023-03-01
install-php-extensions: 3856: !grep: not found

JivinDotL:
fix this issue. !grep => ! grep


mlocati:
Thanks! It's strange that tests didn't catch this error...

PR #716, opened 2023-03-06
Add support for zmq

asamofal:
Hi! Thanks for the project!

Unfortunately, building version `1.1.3` (available on PECL) fail. On GitHub - versions/tags are outdated as well, so I've decided just to use the master branch _(seems like this is the only way to use this extension)_.


mlocati:
:+1:

Issue #633, opened 2022-09-16
Latest Protobuf stable release not working on PHP 8.1


DRoet:
### Version of install-php-extensions

v.1.5.39

### Error description

```sh
#10 506.3 ### INSTALLING REMOTE MODULE protobuf ###
#10 506.3   - Installing protobuf (latest-stable): Downloading (connecting...)Downloading (0%)           Downloading (50%)Downloading (55%)Downloading (100%)
#10 507.0 +-----------------------------------+----------+
#10 507.0 | Package name                      | protobuf |
#10 507.0 | Package version (current release) | 3.18.3   |
#10 507.0 | Package status                    | stable   |
#10 507.0 +-----------------------------------+----------+
#10 513.3 Removing symbols from /usr/local/lib/php/extensions/no-debug-non-zts-20210902/protobuf.so... done (1180488 bytes saved).
#10 513.3 Check if the protobuf module can be loaded... Error loading the "protobuf" extension:
#10 513.4 
#10 513.4 Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetExists($index) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0
#10 513.4 
#10 513.4 Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetGet($index) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0
#10 513.4 
#10 513.4 Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetSet($index, $newval) should either be compatible with ArrayAccess::offsetSet(mixed $offset, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0
#10 513.4 
#10 513.4 Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetUnset($index) should either be compatible with ArrayAccess::offsetUnset(mixed $offset): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0
#10 513.4 
[...]

PR #723, opened 2023-03-24
grpc/protobuf: manually force the latest version

Gounlaf:
Hi :wave:,

This commit follows #633 ; protobuf and grpc are both impacted

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -3067,8 +3067,8 @@ installRemoteModule() {
 				if test $PHP_MAJMIN_VERSION -le 506; then
 					installRemoteModule_version=3.12.4
 				else
-					# see https://github.com/protocolbuffers/protobuf/issues/10619
-					installRemoteModule_version=3.22.1
+					# See https://github.com/protocolbuffers/protobuf/issues/10619
+					installRemoteModule_version=3.22.3
 				fi
 			fi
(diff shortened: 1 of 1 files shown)


mlocati:
Could you amend the commit, by adding a new line containing `Test: grpc, protobuf` (as [described in README](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension))?

Also: this implies that we have to update the script every time a new version of grpc and protobuf are published... Is this something that I'm reclutant to do (maintaining this project it's already rather time consuming)

mlocati:
> there is no guarantee that protoc or grpc release will be published in a "linear" way

The same is valid for any other package published on PECL.

The best solution would be what's described at https://github.com/protocolbuffers/protobuf/issues/10619#issuecomment-1257768192 (and I hope one day PECL will be able to mark a version as the **last** one, even if it has been uploaded before other versions).

mlocati:
@Gounlaf Please remark that, thanks to https://github.com/php/web-pecl/pull/92, we don't need to manually specify the versions of grpc/protobuf anymore (so, I don't have to do stuff like #728, #741, #744, #746, #747, #749, #756, #759, #773, #782, #785, #788, #793 anymore) - see #798

PR #730, opened 2023-04-20
Mark PHP 8.2 as supported by Relay

tillkruss:
There are builds for 8.2 and 8.3 👍


mlocati:
Nope, you don't have to edit the `README.md` file.

As explained [here](https://github.com/mlocati/docker-php-extension-installer#changing-the-supported-php-versions-for-an-already-supported-php-extension), you only need to change the `data/supported-extensions` file (and the `README.md` file will be automatically updated accordingly).

PR #758, opened 2023-06-17
Add support for ext-uv

danog:


mlocati:
@danog I've updated your branch so that we install the beta versions of the extensions (as you can see at https://pecl.php.net/package/uv, there are only beta versions available).

danog:
Yeah I've asked @trowski to tag a stable, he said he'd like to do some more refactoring first, trimming down the extension a bit

PR #765, opened 2023-06-22
Add support for Debian 12 (bookworm) and Alpine 3.18

mlocati:


.github/workflows/monitor-php8.2.yml
```diff
@@ -16,8 +16,10 @@ jobs:
           - alpine3.15
           - alpine3.16
           - alpine3.17
+          - alpine3.18
           - buster
           - bullseye
+          - bookwork
```
mvorisek, line 22:
```suggestion
          - bookworm
```
:)

mlocati:
About memcached: see https://github.com/php-memcached-dev/php-memcached/issues/541

PR #785, opened 2023-08-15
Install grpc 1.57.0 by default

mlocati:


jihadismail8:
i do not understand , why gRPC  and proto is not included in the table of supported extentions ?

jihadismail8:
Yea , I am really sorry , it's never a good idea to work late at night)

PR #809, opened 2023-10-09
feat: add support for predefined processor count

ramunasd:
This helps in situations when `nproc` or other automatic ways gives wrong number of available CPU count.


mlocati:
Thank you!

PR #816, opened 2023-10-14
Drop support for xmldiff on PHP 8.3

mlocati:
See https://github.com/php/pecl-xml-xmldiff/issues/1


nielsdos:
Fixed in the next RC of PHP 8.3.

mlocati:
> Fixed in the next RC of PHP 8.3.

Reminder set: https://github.com/mlocati/docker-php-extension-installer/issues/817

PR #824, opened 2023-10-31
Only install msodbcsql18 on debian 12 Bookworm

ittmann:
Prevents msodbcsql17 being installed


ittmann:
Tested.
Before this change:
```
apt list --installed | grep msodbc

msodbcsql17/now 17.10.5.1-1 amd64 [installed,local]
msodbcsql18/now 18.3.2.1-1 amd64 [installed,local]
```
After:
```
apt list --installed | grep msodbc

msodbcsql18/now 18.3.2.1-1 amd64 [installed,local]
```

mlocati:
Thanks!

PR #848, opened 2023-12-14
Label `org.opencontainers.image.source`

ro0NL:
This should improve compatibility with renovate (dependency updates), and maybe other things :)

Renovate would like source URL + release notes when updating the docker tag in a PR.

See https://github.com/renovatebot/renovate/discussions/26286#discussioncomment-7851042 and  https://docs.renovatebot.com/modules/datasource/docker/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM bash AS build
 
+LABEL org.opencontainers.image.source="https://github.com/mlocati/docker-php-extension-installer"
+
 COPY install-php-extensions /tmp/install-php-extensions
 RUN chmod +x /tmp/install-php-extensions
 


mlocati:
Do we require `org.opencontainers.image.licenses` too?

ro0NL:
Im not entirely familiar with the spec, but AFAIK nothing is required:

https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

mlocati:
I'd add it, just in case.

Something like

```Dockerfile
LABEL org.opencontainers.image.source="https://github.com/mlocati/docker-php-extension-installer" \
      org.opencontainers.image.licenses="MIT"
```

mlocati:
I've just published a new version with this improvement.

ro0NL:
@mlocati it works! https://github.com/renovate-reproductions/26286/pull/1

ro0NL:
@mlocati i also figured out why this worked for our images by default.

We build using github actions.

The `docker/metadata-action` genberates all spec labels:

image

You might want to consider this approach, instead of cluttering the Dockerfile. FYT :)

The renovate integration test is:

```
$ docker inspect mlocati/php-extension-installer:2.1.70 | jq -e '.[].Config.Labels'
{
  "org.opencontainers.image.licenses": "MIT",
  "org.opencontainers.image.source": "https://github.com/mlocati/docker-php-extension-installer"
}
```

PR #879, opened 2024-02-03
Use GitHub mirror instead of VideoLAN's for reliability.

intersectRaven:
VideoLAN's website sometimes comes down at the most inopportune of times and stays down for an unspecified duration. To prevent this from causing errors in build pipelines, I propose we shift it to a much more reliable mirror. Since we don't really rely on latest, this should be fine even if updates on it may lag.

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -2139,7 +2139,7 @@ isLibdav1dInstalled() {
 # Install libdav1d
 installLibdav1d() {
 	printf 'Installing libdav1d\n'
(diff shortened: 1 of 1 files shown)


mlocati:
As [described in README](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension), we need to add `Test: <extension>` to commit messages in order to trigger tests.

For that reason I added 8407525cc745828a71f1d02d95429d6c5275c315 to this pull request.

PR #900, opened 2024-03-18
Add support for imap extension no longer bundled with PHP

mlocati:


mvorisek (approved):
LGTM, Debian 9 is EOL

mlocati:
> Debian 9 is EOL

I didn't port [these lines](https://github.com/mlocati/docker-php-extension-installer/blob/2.2.4/install-php-extensions#L2496-L2505) because the last official PHP version that comes with Debian 9 (stretch) is 7.3: I don't think they'll publish a PHP 8.4 docker image with it.

PR #921, opened 2024-05-14
Prevent problems while installing package

jeff1326:


mlocati:
Could you provide more details about this change? Why is it required?

jeff1326:
@mlocati Unfortunately no. My co-worker did the changes on a older version of this repository. He told me he added these almost everywhere until it works and he don't remember which specific package was causing the issue.

As i see this, this won't break anything at least.

tarlepp:
That `ACCEPT_EULA=Y` sounds like mssql related configuration.

jeff1326:
> That `ACCEPT_EULA=Y` sounds like mssql related configuration.
> 
> https://learn.microsoft.com/en-us/sql/linux/quickstart-install-connect-docker?view=sql-server-ver16&tabs=cli&pivots=cs1-bash#run-the-container-2

That would make sense since we use it in some projet

tarlepp:
although that accept eula thing should not be happen if you're just installing pdo extension for that - not 100% sure about this.

mlocati:
AFAIK the `ACCEPT_EULA` environment variable should be set to install msodbcsql in an unattended way.
But we should already do that: msodbcsql is already being installed without any issue, isn't it?

I won't set a variable nobody knows why and when it should be set...

PR #957, opened 2024-08-08
Remove sqlsrv warning for ARM

mint-jar:
Hi,
Microsoft ODBC 18 supports ARM architecture. It has been tested successfully at our company. The warning is no longer needed.

PR #958, opened 2024-08-09
Remove sqlsrv warning for ARM

mint-jar:
Hi,

Adjusted my changes based on the feedback in: https://github.com/mlocati/docker-php-extension-installer/pull/957

> Microsoft ODBC 18 supports ARM architecture. It has been tested successfully at our company. The warning is no longer needed.


mlocati:
It's not necessary to edit the README: [a GitHub Action](https://github.com/mlocati/docker-php-extension-installer/blob/master/.github/workflows/readme-release.yml) will regenerate it starting from `data/supported-extensions` and `data/special-requirements`).

We only need to edit `data/special-requirements`, like you did here.

Thanks!

PR #884, opened 2024-02-05
Always enable (fast) support for AVIF on Debian 12 (Bookworm)

mlocati:
Let's see if we can use the system-provided libaom, libdav1d, libyuv, and libavif on Debian Bookworm.

If so, we'll always enable AVIF support, since we won't need the time-consuming compilation of those dependencies (and `IPE_GD_WITHOUTAVIF` won't be considered).

PR #986, opened 2024-10-02
Clarify IPE_GD_WITHOUTAVIF behavior

akettmann-id:
Ref: https://github.com/mlocati/docker-php-extension-installer/pull/884

Right now for Debian 12 the envvar IPE_GD_WITHOUTAVIF is ignored. Adding this to the description will save people from having to find it out on their own.


mlocati:
The `-` in the sentence `You can disable AVIF support by setting this environment variable on Debian 11-` means "on debian 11 and previous versions".

But yes, it's rather cryptic...

mlocati:
Superseded by #987

PR #987, opened 2024-10-02
Better explanation of IPE_GD_WITHOUTAVIF

mlocati:
Supersedes #986


akettmann-id:
Thank you!!! This is much clearer!

PR #1009, opened 2024-11-22
Use multi-arch building to have arm64 and amd64 images on docker hub

### LINKED ISSUE #1005: Multi arch images
Although it seems to work fine using this image on arm64, we do get a warning in our pipelines which would be nice to cleanup.

image

Could you publish this image behind a manifest that support different arches please.

baschny:
In order to build a multi arch image (arm64 and amd64) on release on docker hub:

- use the buildx action to load the multi-arch capable builder for docker
- use the `build-push-action` to build and push the multi-arch images (its not possible to do "build" and "push" in separate steps anymore, since the multi-arch build with buildx only supports `--push` and not `--load`).

Tested on my own account:  https://hub.docker.com/r/baschny/php-extension-installer/tags

Resolves  #1005

### DIFF
--- a/.github/workflows/readme-release.yml
+++ b/.github/workflows/readme-release.yml
@@ -170,19 +170,22 @@ jobs:
           username: mlocati
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       -
-        name: Build docker image
-        if: env.VERSIONTAG_THIS != ''
-        run: >
-          docker build
-          --tag "mlocati/php-extension-installer:$VERSIONTAG_THIS"
-          --tag "mlocati/php-extension-installer:${VERSIONTAG_THIS%.*}"
-          --tag "mlocati/php-extension-installer:${VERSIONTAG_THIS%%.*}"
-          --tag mlocati/php-extension-installer:latest
-          .
+        name: Set up docker buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          install: true
       -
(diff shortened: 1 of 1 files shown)


mlocati:
Why just `linux/amd64` and `linux/arm64`?

I'd use all the platforms defined by the official PHP images...

For example, for the latest php:8.3 image I see these architectures at https://hub.docker.com/_/php/tags:

- linux/386
- linux/amd64
- linux/arm/v5
- linux/arm/v7
- linux/arm64/v8
- linux/mips64le
- linux/ppc64le
- linux/s390x

(It seems a bit strange to me that an image cannot be defined as "platform independent", but I guess we have to accept it)

baschny:
Good idea - its so small anyway :) It works: https://hub.docker.com/r/baschny/php-extension-installer/tags, see [a750748](https://github.com/mlocati/docker-php-extension-installer/pull/1009/commits/a750748f1806c981e65e2cc812243b8d9823f97f)

mlocati:
Thanks!

mlocati:
PS: this required a minor fix - see https://github.com/mlocati/docker-php-extension-installer/commit/2654a6902afbf87d616a364cf7eb269360a75de9

mlocati:
And there was a major issue: see #1016 (fixed by 2e59715b3cde9dfbe35be0cea8e509e2fcce5b13)

PR #1010, opened 2024-11-22
Add support for imagick on PHP 8.4

mlocati:


install-php-extensions
```diff
@@ -275,7 +275,9 @@ processPHPModuleArgument() {
 	processPHPModuleArgument_arg="$1"
 	case "$processPHPModuleArgument_arg" in
 		imagick)
-			if test $PHP_MAJMIN_VERSION -ge 803; then
+			if test $PHP_MAJMIN_VERSION -ge 804; then
+				processPHPModuleArgument_arg=imagick/imagick@65e27f2bc02e7e8f1bf64e26e359e42a1331fca1
```
bendavies, line 279: maybe 8.3 can use commit too? which is far more recent than `28f27044e435a2b203e32675e942eb8de620ee58`

bendavies: at least, that's what @mvorisek did here:
https://github.com/mvorisek/image-php/commit/c1ac4c804273cfc2269992123cc107bef4c75634

mvorisek: https://github.com/mvorisek/image-php/blob/master/make.php#L135

mvorisek: yes - it is tested

PR #1029, opened 2024-12-13
relay 8.4 support

tillkruss:
As of [v0.9.1](https://github.com/cachewerk/relay/releases/tag/v0.9.1) we fully support PHP 8.4.


mlocati:
Thanks!

PS: You don't need to manually update the table in the `README.md` file: it's updated automatically starting from the `data/supported-extensions` file: see
https://github.com/mlocati/docker-php-extension-installer/blob/2.7.6/README.md?plain=1#L215-L221

PR #1035, opened 2024-12-17
Fix installing relay on Ubuntu

yatsukhnenko:
Installing issue was found on https://www.drupal.org/project/redis/issues/3493855

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -3837,7 +3837,8 @@ installRemoteModule() {
 						installRemoteModule_distro=alpine3.17
 					fi
 					;;
-				debian)
+				debian | ubuntu)
+					installRemoteModule_distro=debian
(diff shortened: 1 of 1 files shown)


tillkruss:
Relay doesn't have dedicated builds for Ubuntu, because it's the same kernel as debian.

mlocati:
I'm not going to support ubuntu: this project relies on scripts that are only available in official docker images (which only come in Alpine and Debian flavors).

tillkruss:
@mlocati: Looks like the Drupal project is using their own Ubuntu based containers and then use your installer.

mlocati:
> @mlocati: Looks like the Drupal project is using their own Ubuntu based containers and then use your installer.

Any link?

tillkruss:
@berdir can you post a link for @mlocati?

PR #1038, opened 2025-01-03
Add support for mcrypt on PHP 8.4

chrootlogin:
Enables mcrypt on PHP 8.4 because with version 1.0.7 mcrypt is supported.


mlocati:
It seems that mcrypt doesn't support php versions greater than 8.4.0

tarlepp:
> It seems that mcrypt doesn't support php versions greater than 8.4.0

Hmm, isn't that odd - just because php 8.4.0 doesn't even exist first version was 8.4.1

mlocati:
Yep, rather odd. Maybe they meant PHP `< ` 8.4.0, but in the package published on PECL we currently have `<=` 8.4.0

tarlepp:
What I see, it's that people should move out of mcrypt and use eg. openssl instead - sure this is not "easy" way out, but in the long run I think it's the correct path.

chrootlogin:
@tarlepp Thanks for your insight. I'm little bit out of the php coding space. Thats why I thought it will work. I mostly use php for our custom docker images packaging nextcloud etc. 
Now I found out it's an optional dependency of nextcloud. But still creates the question why such big projects are using it.
But anyway I saw mcrypt is not maintained anymore on PECL. So I will close this.

PR #1044, opened 2025-01-09
Add flag to compile Swoole 6.x without io_uring support

MichaelGooden:
This PR adds a flag to optionally compile Swoole without enabling io_uring support.

(I am not sure if this is the best way to implement this in the script)

Security experts generally believe io_uring to be unsafe. In fact Google ChromeOS and Android have turned it off, plus all Google production servers turn it off. Based on the blog published by Google below it seems like a bunch of vulnerabilities related to io_uring can be exploited to breakout of the container. [Google Security Blog]
Other security reaserchers also hold this opinion: see the [Blackhat Presentation] on io_uring exploits.

Impact on IPE:
Secure defaults on many container runtimes and operating systems limit `io_uring` syscalls to privileged users holding `CAP_SYS_ADMIN`. If Swoole is compiled with io_uring support enabled, and the resulting container is run as a non-privileged/non-root user with these secure defaults, the installation will crash with an io_uring permissions error when using any filesystem function hooked by Swoole in a Coroutine.

There does not appear to be a mechanism to disable this behavior at Swoole runtime.

[Google Security Blog]: https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html
[Blackhat Presentation]: https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,10 +449,7 @@ Here's the list of all the supported environment variables:
 | newrelic | `IPE_NEWRELIC_DAEMON=1` | Install the NewRelic daemon  |
 | newrelic | `IPE_NEWRELIC_KEEPLOG=1` | Keep the log files of NewRelic setup (`/tmp/nrinstall-….tar`)  |
 | newrelic | `NR_INSTALL_KEY` | Your New Relic license key |
-| swoole | `IPE_SWOOLE_WITHOUT_IOURING=1` | The io_uring kernel functionality is considered unsafe by security experts [1][2]; By default Swoole 6 and later is configured with io_uring support, use this environment variable to skip io_uring |
-
-[1](https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html)
-[2](https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf)
+| swoole | `IPE_SWOOLE_WITHOUT_IOURING=1` | The io_uring kernel functionality is considered unsafe by security experts (see [here](https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html) and [here](https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf)). By default Swoole 6 and later is configured with io_uring support, use this environment variable to skip configuring io_uring |
 
 ## Special requirements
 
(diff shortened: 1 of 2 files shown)


mlocati:
Thanks!

That's the correct approach (at least, it's the approach used elsewhere in the script).

What about:
1. Disabling iouring by default, inverting the logic (the env var could be something like `IPE_SWOOLE_WITH_IOURING`)
3. Fixing the links in the README
   
   In [the README of your PR](https://github.com/MichaelGooden/docker-php-extension-installer/tree/swoole-without-iouring?tab=readme-ov-file#configuration) we have misplaced links:
   immagine
   I'd replace
   ```md
   | swoole | `IPE_SWOOLE_WITHOUT_IOURING=1` | The io_uring kernel functionality is considered unsafe by security experts [1][2]; By default Swoole 6 and later is configured with io_uring support, use this environment variable to skip io_uring |
   
   [1](https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html)
   [2](https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf)
   ```
   with something like this:
   ```md
   | swoole | `IPE_SWOOLE_WITHOUT_IOURING=1` | The io_uring kernel functionality is considered unsafe by security experts (see [here](https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html) and [here](https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf)). By default Swoole 6 and later is configured with io_uring support, use this environment variable to skip io_uring |
   ```

MichaelGooden:
> 1. Disabling iouring by default, inverting the logic (the env var could be something like `IPE_SWOOLE_WITH_IOURING`)
I have some thoughts on this:
a) Backwards compatibility: there may be users relying on io_uring being enabled by default, per existing behavior
b) This may be an opinionated security position. io_uring is in active development and receives regular security fixes - my understanding is that it isn't inherently dangerous, it just has a very large attack surface and is relatively new code to the kernel

As an end user, I appreciate and rely on the "batteries-included" approach to your installer. My concern would be setting precedent on having to opt-in to otherwise expected defaults.

> 2. Fixing the links in the README
Will be updated shortly.

PR #1046, opened 2025-01-21
Enable Swoole curl-native on Alpine

MichaelGooden:
At minimum Swoole 6 on Alpine doesn't experience the segmentation fault with the native curl implementation, under default installation criteria.

This is the initial attempt at a change, we will want to ensure the test cases for Swoole curl cover this change sufficiently.

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -4109,12 +4109,6 @@ installRemoteModule() {
 					fi
 					;;
 			esac
-			if test $PHP_MAJMIN_VERSION -ge 800; then
-				# see https://github.com/swoole/swoole-src/issues/5365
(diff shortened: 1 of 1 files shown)


mlocati:
@MichaelGooden is this PR ready for review?

MichaelGooden:
I have tested locally that forcing Swoole to use Curl on Debian + PHP 8 fails the test with a segfault.

Based on that I believe the current tests are sufficient for this change in behavior.

MichaelGooden:
> @MichaelGooden is this PR ready for review?

This is now ready for review.

mlocati:
Thanks!

PR #1055, opened 2025-02-05
Test Alpine 3.21

MichaelGooden:
Official PHP images now have an Alpine 3.21 variant, test it.

### DIFF
--- a/.github/workflows/monitor-php8.4.yml
+++ b/.github/workflows/monitor-php8.4.yml
@@ -16,6 +16,7 @@ jobs:
         distro:
           - alpine3.19
           - alpine3.20
+          - alpine3.21
           - bullseye
           - bookworm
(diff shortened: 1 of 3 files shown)


mlocati:
Thanks!

By adding support for Alpine 3.21 we also need to check all the currently supported extensions.

So, before merging this PR, let's wait until the (temporary) PR #1056 finishes (it will try to install all the supported extensions on Alpine 3.21).

mlocati:
@MichaelGooden there's no need for commit ade11f7: [as I wrote above](https://github.com/mlocati/docker-php-extension-installer/pull/1055#issuecomment-2636393494) we are already testing all extensions in #1056 (and I already started fixing the failing ones).

mlocati:
Everything should be ok, thanks!

PR #1080, opened 2025-03-16
Use newer Imagick 3.8.0RC2, with patches for PHP 8.4 and more

thueske:
Hey!

Imagick has released a new version - currently still RC - after a long time. It now also has official PHP 8.4 support. I've been using it for a few days now and it works without any problems.

It is now also included in the official WordPress image:
https://github.com/docker-library/wordpress/blob/89a9728770fd81577a2408254aa007943733152f/latest/php8.4/fpm/Dockerfile#L50C29-L50C74

It would be cool if we had it here too.


mlocati:
I'd wait for a GA version

mlocati:
See #1091

Issue #1084, opened 2025-03-31
Commit archives removed from php-geos extension repo


roncodes:
### Version of install-php-extensions

2.7.28

### Error description

Hi,

Just wanted to report that it seems the php-geos extension has removed the ability to download it's commit archives. This is causing a 404 when attempting to download the php-geos extension. 

I found that there is a more reliable mirror hosted on gihub which has all the same commits/tags/releases etc and simply changing the url and file extension works:

```
https://git.osgeo.org/gitea/geos/php-geos/archive/{commitHash}.tar.gz
to
https://github.com/libgeos/php-geos/archive/{commitHash}.zip
```

Another minor note is there is a more recent hash available for php-geos extension.

### Docker image

dunglas/frankenphp:1.5.0-php8.2-bookworm

### Minimal Dockerfile

```Dockerfile
# syntax = docker/dockerfile:1.2
# Base stage
FROM dunglas/frankenphp:1.5.0-php8.2-bookworm as base

# Install PHP Extensions
RUN install-php-extensions geos
```

PR #1085, opened 2025-04-02
Fix php-geos extension download URL (#1084)

abdomassoun:
## 🛠️ What does this PR do?
This PR updates the download URL for the `php-geos` extension to use the GitHub mirror instead of the old git.osgeo.org source, which is no longer available.

## 🔍 Issue Reference
Fixes #123  

## ✅ Changes Made
- Updated the `php-geos` extension download URL  
- Updated the commit hash for a more recent version


mlocati:
Please read https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension

(We need a commit with a second line containing `Test: geos`)

PR #1098, opened 2025-06-01
Distro Stretch: Added flag --allow-unauthenticated to apt-get install inside installRequiredPackages() function

pavonic:
For archived debian distro Stretch the flag is needed in apt-get install call inside installRequiredPackages() function. Otherwise it fails. 
May be it could be necessary to add the flag in other functions of the srcipt that contains the apt-get install call. But they are for  particular extensions that I didn't use.


mlocati:
#1099 is a better fix IMHO

PR #1129, opened 2025-07-02
Add support for translit

mlocati:


mlocati:
@derickr I keep getting errors like this:

```
Could not download from "https://pecl.php.net/get/translit-0.6.2.tgz", cannot download "pecl/translit" (File https://pecl.php.net:443/get/translit-0.6.2.tgz not valid (received: HTTP/1.1 500 Internal Server Error
  ))
```
See https://github.com/mlocati/docker-php-extension-installer/actions/runs/16029302764/job/45229860754?pr=1129#step:3:97

This doesn't always occur.

We used to have these errors with the old pecl server, but with the new one we see them much often... I had to run tests 5 times and I still have to run them the 6th time...

derickr:
Hmm, `pecl install translit-0.6.2` produces the following error in the log:
```
[Wed Jul 02 17:26:30.365840 2025] [php:error] [pid 197630:tid 197630] [client 109.146.183.231:0] PHP Fatal error:  Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'release' cannot be null in /var/www/sites/pecl.php.net/src/Database.php:53\nStack trace:\n#0 /var/www/sites/pecl.php.net/src/Database.php(53): PDOStatement->execute()\n#1 /var/www/sites/pecl.php.net/src/Release.php(594): App\\Database->run()\n#2 /var/www/sites/pecl.php.net/src/Release.php(539): App\\Release->logDownload()\n#3 /var/www/sites/pecl.php.net/public_html/get(79): App\\Release->HTTPdownload()\n#4 {main}\n  thrown in /var/www/sites/pecl.php.net/src/Database.php on line 53
```

I think it's a problem with this specific release (only)? 

I have deleted it, as there are newer versions too. 

Why are you trying to install that old release?

install-php-extensions
```diff
@@ -4545,6 +4545,17 @@ installRemoteModule() {
 			mv "$installRemoteModule_src" $(getPHPExtensionsDir)/tideways.so
 			installRemoteModule_manuallyInstalled=1
 			;;
+		translit)
+			if test -z "$installRemoteModule_version"; then
+				if test $PHP_MAJMIN_VERSION -lt 506; then
+					installRemoteModule_version=0.6.2
```
derickr, line 4551: You probably want to change this to `0.6.1` for `< 5.6`

mlocati: Thanks!

mlocati:
> Why are you trying to install that old release?

Just because we can (or at least, we should be able to do according to the specs 😜).

Issue #1138, opened 2025-08-13
`install-php-extensions imagick` fails


kwarcu:
### Version of install-php-extensions

install-php-extensions v.2.8.6

### Error description

`install-php-extensions imagick` fails when trying to run on `php:8.4-fpm` container

```
install-php-extensions imagick
install-php-extensions v.2.8.6
#StandWithUkraine
Updating channel "pecl.php.net"
Channel "pecl.php.net" is up to date
Get:1 http://deb.debian.org/debian trixie InRelease [138 kB]
Get:2 http://deb.debian.org/debian trixie-updates InRelease [47.1 kB]
Get:3 http://deb.debian.org/debian-security trixie-security InRelease [43.4 kB]
Get:4 http://deb.debian.org/debian trixie/main arm64 Packages [9604 kB]
Get:5 http://deb.debian.org/debian trixie-updates/main arm64 Packages [2432 B]
Get:6 http://deb.debian.org/debian-security trixie-security/main arm64 Packages [5692 B]
Fetched 9840 kB in 3s (3236 kB/s)
Reading package lists...
FAILED TO LIST THE WHOLE PACKAGE LIST FOR
^libmagickwand-6.q16-[0-9]+$ ^libmagickcore-6.q16-[0-9]+-extra$

COMMAND OUTPUT:
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package ^libmagickwand-6.q16-[0-9]+$
E: Couldn't find any package by glob '^libmagickwand-6.q16-[0-9]+$'
E: Couldn't find any package by regex '^libmagickwand-6.q16-[0-9]+$'
E: Unable to locate package ^libmagickcore-6.q16-[0-9]+-extra$
E: Couldn't find any package by glob '^libmagickcore-6.q16-[0-9]+-extra$'
E: Couldn't find any package by regex '^libmagickcore-6.q16-[0-9]+-extra$'

E: apt-get failed
```

### Docker image

php:8.4-fpm

### Minimal Dockerfile

```Dockerfile
FROM php:8.4-fpm

COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
RUN install-php-extensions mysqli gd zip pdo_mysql opcache pcntl bcmath imagick
```

nathanpurcell:
I was just about to open an issue also - same problem.

kwarcu:
Probably caused by the fact that apt already installs libs in v7 not 6

[...]

PR #1139, opened 2025-08-13
Fix 1138 - new imagick library version

kwarcu:
Fix https://github.com/mlocati/docker-php-extension-installer/issues/1138

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1005,7 +1005,7 @@ buildRequiredPackageLists() {
 				fi
 				;;
 			imagick@debian)
-				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent ^libmagickwand-6.q16-[0-9]+$ ^libmagickcore-6.q16-[0-9]+-extra$"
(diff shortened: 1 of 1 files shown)


mlocati:
I don't think that's the correct way to fix it.

In any case, I'm going to check all the PHP extensions with Debian Trixie - see #1140

mlocati:
Also: when submitting a pull request that affects a PHP extension, please follow [these guidelines](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension) (otherwise the tests won't check if the change breaks something).

kwarcu:
> I don't think that's the correct way to fix it.
> 
> In any case, I'm going to check all the PHP extensions with Debian Trixie - see #1140

Thanks! I was just trying to act quickly to find walkaround.

>Also: when submitting a pull request that affects a PHP extension, please follow [these guidelines](https://github.com/mlocati/docker-php-extension-installer#improving-code-for-an-already-supported-extension) (otherwise the tests won't check if the change breaks something).

Will do. 

All the best & thanks for your work!

PR #1170, opened 2025-10-01
Add support for PHP 8.5 RC

mlocati:


mlocati:
The failure of `intl` on php:8.1-cli-alpine3.22 has been fixed by #1171

PR #1193, opened 2025-11-24
Add support for solr on PHP 8.5

mlocati:


mlocati:
See https://github.com/php/pecl-search_engine-solr/issues/94

PR #1201, opened 2025-12-05
Add support for xdebug on PHP 8.5

mlocati:


leonboot:
Let's hope Xdebug 3.5.0 ends up on PECL 🤞 Seeing PECL is considered deprecated I hope it will!

philbates35:
> Let's hope Xdebug 3.5.0 ends up on PECL 🤞 Seeing PECL is considered deprecated I hope it will!

The wording of the XDebug 3.5.0 [release notes](https://xdebug.org/announcements/2025-12-04) are quite specific in that its only been released on PIE and not PECL (unlike the 3.4.7 [release notes](https://xdebug.org/announcements/2025-10-26)), so its possible that its intentional that it isn't coming to PECL...

jdreesen:
See: https://bugs.xdebug.org/view.php?id=2388#c7427

PR #1202, opened 2025-12-06
Add PHP version support for swoole 8.5

francoism90:
It seems this is now supported upstream.


francoism90:
Hmm: https://github.com/swoole/swoole-src/blob/384f9fdc212e49314246afe9a87d3b7452f9b493/package.xml#L2827

Why does https://pecl.php.net/package/swoole/6.1.4 still report PHP 8.4.99?

mlocati:
> Why does https://pecl.php.net/package/swoole/6.1.4 still report PHP 8.4.99?

See https://github.com/swoole/swoole-src/pull/5944#issuecomment-3631328058

francoism90:
I'll keep an eye on it. :)

PR #1216, opened 2025-12-24
Add support for blackfire on PHP 8.5

Teko012:
It supports 8.5: https://blog.blackfire.io/blackfire-now-supports-php-8-5.html


mlocati:
Thanks!

Issue #1189, opened 2025-11-19
Add PHP 8.5 patch for amqp


shyim:
Hey,

could we adjust the extension installer to automatically patch https://github.com/php-amqp/php-amqp/pull/595 so amqp also works with PHP 8.5?

mlocati:
You can tell `install-php-extensions` to install amqp directly from github, specifying the hash of the [latest commit](https://github.com/php-amqp/php-amqp/pull/595/commits/977449987412a3d5c59a036dbab8b6d67764bb3e) of that pull request:

```sh
install-php-extensions php-amqp/php-amqp@977449987412a3d5c59a036dbab8b6d67764bb3e
```

shyim:
Thank you!

PR #1219, opened 2026-01-05
Add PHP version support for amqp 8.5

nlzet:
I believe amqp should now be supported on 8.5

Relates to https://github.com/mlocati/docker-php-extension-installer/issues/1189


mlocati:
Thanks!

PR #1223, opened 2026-01-12
Download pdo_snowflake from codeload.github.com

mlocati:


mlocati:
About commit 266e527fb69ee180b7f9d2dfa7244c508a2cdb88: since pdo_snowflake requires PHP 8.1+ and [there's no official jessie/stretch image with PHP 8.1+](https://github.com/mlocati/docker-php-images-matrix?tab=readme-ov-file#the-matrix), such special requirements are pretty useless.

PR #1234, opened 2026-02-11
Be lazy about the syntax of package.xml files

### LINKED ISSUE #1233: Version from `package.xml` is not parsed from v1.0
### Version of install-php-extensions

2.9.29

### Error description

According https://pear.php.net/manual/en/developers.packagedef.intro.php the version is located under `release` node in version 1.0.

But the current version of this installed expects is somewhere else - https://github.com/mlocati/docker-php-extension-installer/blob/2.9.29/install-php-extensions#L241 .

To fix this issue either `/{$ns}package/{$ns}release/{$ns}version` query should be used or used as fallback (I am not sure if the the current query is used in any package nor if allowed by v1.0 spec).

### Docker image

php:8.5

mlocati:
Close #1233

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -234,13 +234,20 @@ if ($xpath->query('/v20:package/v20:dependencies')->length === 1) {
 	$ns = '';
 } else {
 	fwrite(STDERR, "Unsupported namespace of the XML of package version details\n");
+	exit(1);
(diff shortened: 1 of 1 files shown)


mlocati:
PS: pull request #1234... nice number 😜

PR #1236, opened 2026-02-25
Optimize configureInstaller() to use PECL only when needed

taai:
No need to fetch any extension lists from PECL when installing bundled extensions.


jonnyeom:
I really like this!

PR #1237, opened 2026-03-02
sqlsrv and pdo_sqlsrv: enforce version 5.12.0 for php versions lower then 8.3

goestijn:


mlocati:
Thanks! #1239 is a better PR imho...,

PR #1247, opened 2026-03-10
Fix libmemcachedutil2 on debian 13

HB9HIL:
The package was renamed in debian 13. Fails for armv7 builds

See here: https://packages.debian.org/search?searchon=names&keywords=libmemcachedutil2

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1152,7 +1152,11 @@ buildRequiredPackageLists() {
 				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libmemcached-dev zlib-dev"
 				;;
 			memcached@debian)
-				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent libmemcachedutil2"
+				if test $DISTRO_VERSION_NUMBER -ge 13; then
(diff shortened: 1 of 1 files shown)


mlocati:
What about 32 bit systems?

HB9HIL:
Should not be a problem. I did a bit of a deeper research and `libmemcachedutil2t64` should also work on 32 bit systems without any issues. I can't test "real" 32bit hardware but I made some testbuilds of our project on i386 which just worked fine with the modified script. In theory it should work to just replace `libmemcachedutil2` with `libmemcachedutil2t64` also on bookworm, but it worked so beautiful and stable on bookworm I thought it's a good idea to keep that behaviour and just catch trixie which does not provide `libmemcachedutil2` anymore. 

Did you noticed some specific issues?

mlocati:
I thought that the trailing "64" was somehow related to the bitness of the package, but it's only related to the bitness of time_t (and it's available on 32bit systems too - see https://packages.debian.org/trixie/i386/libmemcachedutil2t64/filelist ).

So, yep, it seems its safe to merge this PR.

Thanks!

HB9HIL:
Oh yeah sorry I coul dhave provide that information haha :D thx for merge :)

PR #1252, opened 2026-03-11
Fix for the Swoole Firebird on Alpine arm64

revenkroz:
Fix after https://github.com/mlocati/docker-php-extension-installer/pull/1246

We need to disable Firebird support for `swoole` on `aarch64-alpine-linux-musl` as current `install-php-extensions`  pulls in a Firebird 5 source build. When installing `swoole` on:
  - Alpine
  - arm64 / aarch64
  - PHP 8.4+

Firebird build fails during linking with errors like:

```text
undefined reference to `_mcount'
collect2: error: ld returned 1 exit status
```

**A minimal repro is:**
```
FROM php:8.4-cli-alpine
COPY --from=ghcr.io/mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
RUN install-php-extensions swoole
```

As a result, install-php-extensions swoole fails on Alpine arm64 even if Firebird support is not required by the user. 

The fix only affects `swoole + Alpine + arm64/aarch64 musl` combo and keeps Firebird enabled elsewhere.

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1513,7 +1513,7 @@ buildRequiredPackageLists() {
 						if test $PHP_MAJMIN_VERSION -ge 802; then
 							buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent libssh2"
 							buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile ncurses-dev libssh2-dev"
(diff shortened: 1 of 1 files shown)


mlocati:
I'd instead fix compiling firebird on arm64 - see #1254

PR #1274, opened 2026-04-16
Improve Relay Install (for libssl3 / zts)

gobalopper:
Improve Relay extension by adding a check if ZTS should be used and a better check for libssl3.

Test: relay

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -62,7 +62,7 @@ fi
 
 # Reset the Internal Field Separator
 resetIFS() {
-	IFS='	 
+	IFS='
 '
 }
 
@@ -3036,7 +3036,7 @@ installBundledModule() {
 +  PHP_DBA_DB_CHECK(4, db-5.3 db-5.1 db-5.0 db-4.8 db-4.7 db-4.6 db-4.5 db-4.4 db-4.3 db-4.2 db-4.1 db-4.0 db-4 db4 db, [(void)db_create((DB**)0, (DB_ENV*)0, 0)])
  fi
  PHP_DBA_STD_RESULT(db4,Berkeley DB4)
- 
+
 EOF
 			fi
 			docker-php-ext-configure dba --with-db4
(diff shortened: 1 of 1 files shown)


install-php-extensions
```diff
@@ -62,7 +62,7 @@ fi
 
 # Reset the Internal Field Separator
 resetIFS() {
-	IFS='	 
+	IFS='
```
mlocati: Please don't youch this line

install-php-extensions
```diff
@@ -3036,7 +3036,7 @@ installBundledModule() {
 +  PHP_DBA_DB_CHECK(4, db-5.3 db-5.1 db-5.0 db-4.8 db-4.7 db-4.6 db-4.5 db-4.4 db-4.3 db-4.2 db-4.1 db-4.0 db-4 db4 db, [(void)db_create((DB**)0, (DB_ENV*)0, 0)])
  fi
  PHP_DBA_STD_RESULT(db4,Berkeley DB4)
- 
+
```
mlocati: Please revert this change

install-php-extensions
```diff
@@ -3122,7 +3122,7 @@ EOF
 						patch /usr/src/php/ext/intl/config.m4 <<'EOF'
 @@ -85,7 +85,16 @@ if test "$PHP_INTL" != "no"; then
      breakiterator/codepointiterator_methods.cpp"
- 
+
```
mlocati: Please revert this change

gobalopper:
I should point out this is part of trying to get Relay to work with FrankenPHP using a Debian docker image. This doesn't fix the underlying problem but should make things work once Relay fixes another issue on their side.

install-php-extensions
```diff
@@ -4282,10 +4282,8 @@ installRemoteModule() {
 					fi
 					;;
 				debian)
-					case "$(dpkg -l 'libssl*' | grep -E '^ii ' | cut -d' ' -f3)" in
```
mlocati, line 4285: What was wrong with the previous code?

gobalopper: If, for example, you had `libssl-dev` installed it would find two packages and the match wouldn't work.

mlocati:
Thanks!

BWJesseR:
@mlocati question about the tests. I am trying to understand how it works when a package has both NTS and ZTS versions, does it test both?

mlocati:
> @mlocati question about the tests. I am trying to understand how it works when a package has both NTS and ZTS versions, does it test both?

Nope, we just run tests against the "default" CLI docker images (they are already so many that sometimes tests take more than 1 hour)

PR #1275, opened 2026-04-20
imagick: install libheif-plugin-aomenc on Debian 13+ for AVIF support

lenlorijn:
On Debian 13 the `libheif1` codecs were split into separate plugin packages, so `imagick` images no longer include an AV1 encoder and ImageMagick can't write AVIF. This adds `libheif-plugin-aomenc` on Debian 13+; bookworm and Alpine are unaffected afaik.

Test: imagick

### DIFF
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -1071,7 +1071,7 @@ buildRequiredPackageLists() {
 				;;
 			imagick@debian)
 				if test $DISTRO_VERSION_NUMBER -ge 13; then
-					buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent ^libmagickwand-7.q16-[0-9]+$ ^libmagickcore-7.q16-[0-9]+-extra$"
+					buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent ^libmagickwand-7.q16-[0-9]+$ ^libmagickcore-7.q16-[0-9]+-extra$ libheif-plugin-aomenc"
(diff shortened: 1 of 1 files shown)


mlocati:
Do you know any quick way to check if imagick can write AVIF? [It seems](https://github.com/mlocati/docker-php-extension-installer/pull/1276) that `Imagick::queryFormats()` includes `'AVIF'` even on debian 13+ without `libheif-plugin-aomenc`...

lenlorijn:
Yes no problem. The platform check we use uses the following code:
```php
try {
    $imagick = new Imagick();
    $imagick->newImage(10, 10, new ImagickPixel('red'));
    $imagick->setImageFormat('avif');
    $writeBytes = strlen($imagick->getImageBlob());
    $imagick->clear();
} catch (\Throwable $e) {
    $writeError = $e->getMessage();
}
```

The version string should also have changed I believe. 
```php
$versionString = (new Imagick())->getVersion()['versionString'] ?? '';
$hasAom = stripos($versionString, 'aom') !== false;
```

mlocati:
Thanks!

mlocati:
In #1277 we also check for write support, and indeed this #1275 fixes the issue: compare
- https://github.com/mlocati/docker-php-extension-installer/actions/runs/24777040860 (before this fix)
- https://github.com/mlocati/docker-php-extension-installer/actions/runs/24789882497 (after this fix)

PR #1290, opened 2026-05-23
Install ddtrace via datadog-setup.php on PHP 7.0+

### LINKED ISSUE #1163: Datadog Profiler module
The [datadog profiler component](https://github.com/DataDog/dd-trace-php/tree/master/profiling) is a sibling to the ddtrace component that's already in this helper library.

The full doco for PHP integration is here: https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/php/

Specifically, `--enable-profiling` (`datadog-profiling` extension) turns this on, though I'm not sure how it's intended to be used from the [PECL package](https://pecl.php.net/package/datadog_trace).

- https://github.com/DataDog/dd-trace-php
- https://github.com/DataDog/dd-trace-php/blob/master/profiling/datadog-profiling-install.sh

Not sure if it's worth [doing appsec](https://github.com/DataDog/dd-trace-php/tree/master/appsec) at the same time?

Teko012:
On PHP 7.0+ this PR switches the ddtrace installation path from the existing PECL + Cargo build to Datadog's official installer, `datadog-setup.php`. The PECL/Cargo path is preserved for PHP <7.0 and as a fallback when a user pins ddtrace to a version below 1.0.0.

### Why

- `datadog-setup.php` is Datadog's supported install path for the PHP tracer. Using it keeps the installation aligned with Datadog's release artifacts and version-pinning rules instead of relying on PECL's mirror of `datadog/dd-trace`.
- It can co-install the companion modules (`ddappsec`, `datadog-profiling`) in lockstep with the matching tracer build, which is awkward to do via PECL.
- It avoids building Rust from source via Cargo for the common case, which currently pulls `curl-dev automake libtool` (Alpine) or `libssl libcurl4-openssl-dev automake libtool` (Debian) and adds noticeable build time/image size.

### What changes

**`installRemoteModule()` — the `ddtrace)` branch (`install-php-extensions`)**

The selection logic is reorganized around the PHP version:

- **PHP <7.0** — unchanged path: build from source via PECL + `installCargo`. If a user explicitly requests `ddtrace-1.0.0` or newer on PHP <7.0, the script now errors out (`ddtrace <version> requires PHP 7.0 or later`) instead of silently failing later. With no version supplied, it still defaults to `0.75.0` (newer 0.x releases refuse PHP 5 in their `configure` script).
- **PHP 7.0+ with an explicit pre-1.0.0 version** — keep using the legacy `installCargo` path so older 0.x releases stay reproducible.
- **PHP 7.0+, default or 1.0.0+** — call the new `installDDTrace` helper and set `installRemoteModule_manuallyInstalled=2` so the surrounding code skips the standard PECL pipeline (this is the same convention used for other manually-installed modules like blackfire).

PECL stability flags (`stable`, `beta`, …) are still resolved via `resolvePeclStabilityVersion` before any of the branches above run.

**New `installDDTrace()` helper (`install-php-extensions`)**

```sh
installDDTrace [version]
```

- Downloads `datadog-setup.php` from the corresponding GitHub release:
  - no version → `https://github.com/DataDog/dd-trace-php/releases/latest/download/datadog-setup.php`
  - version `X.Y.Z` → `https://github.com/DataDog/dd-trace-php/releases/download/X.Y.Z/datadog-setup.php`
- Runs it as `php datadog-setup.php --php-bin=php [extra args]`.
- Honors two new environment variables (see below) to optionally append `--enable-appsec` and/or `--enable-profiling`.
- Uses `curl $IPE_CURL_FLAGS -sSLf` for the download so the existing `IPE_INSECURE=1` toggle keeps working, and errors out cleanly with a `### ERROR ### Failed to download …` message if the download fails.
- Cleans up the temporary `datadog-setup.php` once the install completes.

**New environment variables (`README.md`)**

| Variable | Effect |
| --- | --- |
| `IPE_DD_APPSEC=1` | Pass `--enable-appsec` to `datadog-setup.php`, installing the `ddappsec` module alongside ddtrace. Requires PHP 7.0+ and the 1.x tracer line. |
| `IPE_DD_PROFILING=1` | Pass `--enable-profiling` to `datadog-setup.php`, installing the `datadog-profiling` module. Requires PHP 7.1+ on Alpine 3.13+ (any Debian works) and the 1.x tracer line. |

Both follow the existing `IPE_*` env-var conventions (accept `1`, `y…`, `Y…`).

**`buildRequiredPackageLists()` — `ddtrace@alpine` and `ddtrace@debian` (`install-php-extensions`)**

The development packages that were unconditionally pulled in for ddtrace are now gated behind `PHP_MAJMIN_VERSION -lt 700`, because they're only needed by the Cargo build path:

- Alpine: `curl-dev automake libtool` — installed only on PHP <7.0.
- Debian: `$buildRequiredPackageLists_libssl libcurl4-openssl-dev automake libtool` — installed only on PHP <7.0.

For PHP 7.0+ the `datadog-setup.php` flow installs the prebuilt tracer binary, so none of these dev packages are needed.

**Test script: `scripts/tests/ddtrace`**

Adds an executable PHP test that:
1. Verifies `ddtrace_version()` is defined and prints the resolved version.
2. If `IPE_DD_PROFILING` was set during install, verifies the `datadog-profiling` extension is loaded.
3. If `IPE_DD_APPSEC` was set during install, verifies the `ddappsec` extension is loaded.

The script exits non-zero on any failure so it can be wired into the existing extension test matrix.

### Backwards compatibility

- `install-php-extensions ddtrace` on PHP <7.0 behaves exactly as before (default `0.75.0`, Cargo build, same build deps).
- `install-php-extensions ddtrace-<0.x.y>` on PHP 7.0+ keeps the previous Cargo-based path.
- The only behavioral change for existing users on PHP 7.0+ without a pinned version is the switch from the PECL/Cargo build to the official `datadog-setup.php` install — which yields the same `ddtrace` extension but no longer leaves `curl-dev`/`libcurl4-openssl-dev`/`automake`/`libtool` lingering as runtime-installed packages.

Closes #1163.

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,8 @@ Here's the list of all the supported environment variables:
 | newrelic | `IPE_NEWRELIC_DAEMON=1` | Install the NewRelic daemon  |
 | newrelic | `IPE_NEWRELIC_KEEPLOG=1` | Keep the log files of NewRelic setup (`/tmp/nrinstall-….tar`)  |
 | newrelic | `NR_INSTALL_KEY` | Your New Relic license key |
+| ddtrace | `IPE_DD_APPSEC=1` | Also enable the Datadog [Application Security Monitoring](https://docs.datadoghq.com/security/application_security/) module (`ddappsec`). Requires PHP 7.0+ and the 1.x tracer line |
+| ddtrace | `IPE_DD_PROFILING=1` | Also enable the Datadog [Continuous Profiler](https://docs.datadoghq.com/profiler/) module (`datadog-profiling`). Requires PHP 7.1+, Alpine 3.13+ (or any Debian) and the 1.x tracer line |
 | swoole | `IPE_SWOOLE_WITHOUT_IOURING=1` | The io_uring kernel functionality is considered unsafe by security experts (see [here](https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html) and [here](https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf)). By default Swoole 6 and later is configured with io_uring support, use this environment variable to skip configuring io_uring |
 | saxon | `IPE_SAXON_EDITION=EE` | The Saxon edition to be used: `EE` for Enterprise Edition (default), `PE` for Professional Edition, `HE` for Home Edition |
 
--- a/install-php-extensions
+++ b/install-php-extensions
@@ -825,11 +825,15 @@ buildRequiredPackageLists() {
 				;;
 			ddtrace@alpine)
 				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent libgcc"
-				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile curl-dev automake libtool"
+				if test $PHP_MAJMIN_VERSION -lt 700; then
+					buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile curl-dev automake libtool"
+				fi
 				;;
 			ddtrace@debian)
-				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent $buildRequiredPackageLists_libssl"
-				buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libcurl4-openssl-dev automake libtool"
+				if test $PHP_MAJMIN_VERSION -lt 700; then
+					buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent $buildRequiredPackageLists_libssl"
+					buildRequiredPackageLists_volatile="$buildRequiredPackageLists_volatile libcurl4-openssl-dev automake libtool"
+				fi
 				;;
 			dba@alpine)
 				buildRequiredPackageLists_persistent="$buildRequiredPackageLists_persistent db"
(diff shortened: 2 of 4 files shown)


Teko012:
@mlocati What do you think about this?

mlocati:
That's great

mlocati:
Could you add a commit whose message follows the rules described here? https://github.com/mlocati/docker-php-extension-installer/#improving-code-for-an-already-supported-extension

That way the CI will test installing ddtrace

Teko012:
> Could you add a commit whose message follows the rules described here? https://github.com/mlocati/docker-php-extension-installer/#improving-code-for-an-already-supported-extension
> 
> That way the CI will test installing ddtrace

@mlocati Yes, added it. I hope it's good.

mlocati:
> Yes, added it. I hope it's good.

Yep, the tests are being executed now 👍 (but they fail)

Teko012:
> > Yes, added it. I hope it's good.
> 
> Yep, the tests are being executed now 👍 (but they fail)

Oh no, let me investigate

Teko012:
@mlocati Fixed it, thank you for letting me know! Let me know what you think.

Teko012:
@mlocati What do you think about it? Can we merge it?

mlocati:
That's great!

PR #1293, opened 2026-05-28
Add support for parallax

AndreyMashukov:
This adds support for installing the `parallax` extension.

**About the extension** — [php-parallax](https://github.com/AndreyMashukov/php-parallax) is a Go/Rust-semantic concurrency primitive for PHP: every spawned task runs on its own OS thread inside an isolated PHP runtime, captures are deep-cloned at spawn time (Rust `move` semantics), and one failing task never affects its siblings.

**Hard requirements**: PHP 8.4+ and a ZTS build (`--enable-zts`). The `special-requirements` line marks the ZTS dependency the same way `parallel` and `pthreads` do; the supported-versions line only lists 8.4 and 8.5.

**Install path**: the PECL submission is queued but not yet uploaded, so the install hook builds from the tagged GitHub tarball via `phpize`. The shape mirrors the existing `cassandra)` case (tarball → phpize → configure → make install → mark `installRemoteModule_manuallyInstalled=1`).

Verified locally on `php:8.4-zts-alpine` and `php:8.4-zts-bookworm`:

```
docker run --rm -v "\$PWD/install-php-extensions:/usr/local/bin/install-php-extensions" \
    php:8.4-zts-alpine sh -c 'chmod +x /usr/local/bin/install-php-extensions && install-php-extensions parallax && php -m | grep parallax'
```

Once PECL accepts the package I'll follow up with a smaller PR that switches to the standard \`pecl install\` path and drops the custom build hook.


mlocati:
You can install that extension with pie

PR #1296, opened 2026-06-04
Add support for v8js

### LINKED ISSUE #729: Support for v8js extension
Hi,

Firstly, thanks for this great repo! Could you please share information about the `v8js` extension? Does the repo has support for this extension? If not exists, could you please add it?

Thanks.

standa:
Builds v8js from the phpv8/v8js GitHub `php8` branch (PECL has no PHP 8.x release) and links it against the V8 engine shipped by the distro's Node.js packages (nodejs-libs on Alpine, libnode-dev on Debian), as discussed in #729.

Tested on alpine 3.19/3.22/3.23 and debian bookworm/trixie. Older distros are excluded via special-requirements either because they lack the nodejs-libs package (alpine <= 3.18) or because their bundled V8 is < 9.0 (buster, bullseye).

Closes #729

Test: v8js


data/special-requirements
```diff
@@ -16,6 +16,7 @@ simdjson !jessie !stretch
 sodium !jessie
 sqlsrv !7.1-alpine3.9 !7.1-alpine3.10
 statgrab !alpine3.12 !alpine3.13 !alpine3.14 !alpine3.15 !alpine3.16
+v8js !buster !bullseye !alpine3.9 !alpine3.10 !alpine3.11 !alpine3.12 !alpine3.13 !alpine3.14 !alpine3.15 !alpine3.16 !alpine3.17 !alpine3.18
```
mlocati: The oldest Alpine version that comes with PHP 8.0 is 3,12; there's no need to list the older ones - see https://github.com/mlocati/docker-php-images-matrix#the-matrix

standa: Fixed. Thanks!

mlocati:
Thanks!

PR #1298, opened 2026-06-05
supports mjml extension

guillaume-sainthillier:
This PR install php rust extension from https://github.com/alekitto/mjml-php


mlocati:
You can use pie to install that extension

PR #1294, opened 2026-06-01
Fix SaxonC URL for version >=12.6

benito103e:
Starting with SaxonC v12.6, Saxonica renamed their release archives. Passing an explicit version >= 12.6 to `install-php-extensions saxon-<version>` results in a 404 because the legacy URL pattern no longer exists on their server.

**Before v12.6** :
  https://downloads.saxonica.com/SaxonC/HE/12/libsaxon-HEC-linux-x86_64-v12.5.zip

**Since v12.6**:
  https://downloads.saxonica.com/SaxonC/HE/12/SaxonCHE-linux-x86_64-12-6-0.zip

Two things changed:
  1. Archive prefix: `libsaxon-${edition}C-` → `SaxonC${edition}-`
  2. Version format: `v12.6` (dotted, prefixed) → `12-6-0` 

Official listing: https://downloads.saxonica.com/SaxonC/HE/12/index.html

PR #1299, opened 2026-06-05
Fix version 12.3, 12,4, 12.6-12.9 installation for SaxonC

benbrummer:
With the release of SaxonC 13, which currently fails to compile it was not possible to install any version above 12.x > 12.5

Urls for 12.0+12.1 are correct now (compilation fails)
Urls for 0, 1, 11 should work (compilation not tested)


mlocati:
I also tried to fix saxon, but there's something broken - see my report at https://saxonica.plan.io/issues/7113

benbrummer:
I did not check for the already existing PR https://github.com/mlocati/docker-php-extension-installer/pull/1294

mlocati:
Superseded by #1300

PR #1300, opened 2026-06-08
Fix installing Saxon

mlocati:
- see https://saxonica.plan.io/issues/7113 for a discussion about installation of Saxon 13
- supersedes #1294 (thanks @benito103e - but, since I'm Italian, no thanks for your GitHub nickname 😉)
- supersedes #1299 (thanks @benbrummer)


benito103e:
Thanks for your fix.
;) sorry for my nickname ! It was a mistake of youth and lack of knowledge :) !

PR #1304, opened 2026-06-10
Add Ghostscript to persistent package list for ImageMagick

Teko012:
This pull request makes a small update to the `install-php-extensions` script to improve dependency handling for ImageMagick extensions.

- Added `ghostscript` to the list of persistent required packages when installing ImageMagick-related PHP extensions, ensuring better compatibility and support for image processing features.

The Alpine variant already has it, but it's missing on Debian. It is required for PDF handling.

### DIFF
--- a/scripts/tests/imagick
+++ b/scripts/tests/imagick
@@ -5,13 +5,6 @@ require_once __DIR__ . '/_bootstrap.php';
 
 list($os, $osVersion) = inspectOS();
 
-// For every format we require, define how to obtain a test sample so we can
-// actually exercise the format instead of just trusting Imagick::queryFormats():
-//   true     => let Imagick encode the sample itself, then read it back
-//               (round-trip: exercises both the encoder and the decoder).
-//   <string> => a literal sample already in that format, which we only read back
-//               (exercises the decoder; used for formats Imagick can't reliably
-//               write, but whose reading relies on an external delegate).
 $requiredFormats = [
     'BMP' => true,
     'GIF' => true,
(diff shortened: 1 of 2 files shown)


mlocati:
How big is ghostscript?

Teko012:
@mlocati On Alpine, which the script already installs, it's 62.9MiB. The one we add for Debian now is 57.5 MB.

mlocati:
Do you have a small PHP script that can be used to test whether imagick works correctly with PDF files?

Teko012:
@mlocati I added a script. The test seems to have failed due to some unrelated Docker issue, but the relevant part has passed. Maybe you can rerun the failed ones, they should pass too.

Teko012:
@mlocati Can we merge this in? It might cause a merge conflict with https://github.com/mlocati/docker-php-extension-installer/pull/1306, which I would need to fix

scripts/tests/imagick
```diff
@@ -5,14 +5,22 @@ require_once __DIR__ . '/_bootstrap.php';
...
+//   true     => let Imagick encode the sample itself, then read it back
+//               (round-trip: exercises both the encoder and the decoder).
+//   <string> => a literal sample already in that format, which we only read back
+//               (exercises the decoder; used for formats Imagick can't reliably
+//               write, but whose reading relies on an external delegate).
 $requiredFormats = [
     'BMP' => true,
     'GIF' => true,
     'JPG' => true,
-    'PDF' => false,
```
mlocati, line 12: You removed testing the case when imagick supports reading a format, but doesn't support writing it (eg when the value is `false`).
For example, we previously had `'PDF' => false`, meaning that imagick was listed in the result of `Imagick::queryFormats()`.

Also, please don't be so verbose in comments.

Teko012: Adjusted it. Can you check it now?

Teko012:
@mlocati WDYT? Ready to merge?

Teko012:
Looks like we don't have a merge conflict.

Teko012:
@mlocati Side note: I ran all the tests successfully on all extensions. It looks like no other issues remain, so this is the only thing left. Can we release this too?

mlocati:
Thanks!

PR #1307, opened 2026-06-16
Fix wikidiff2 on php:8.1

Teko012:
Fix wikidiff2 on php:8.1

Example failure is [here](https://github.com/mlocati/docker-php-extension-installer/actions/runs/27313201250/job/80687936709)


mlocati:
Thanks!

PR #1315, opened 2026-08-21
Update cassandra

maxm86545:
- Switch the `cassandra` extension to [he4rt/scylladb-php-driver](https://github.com/he4rt/scylladb-php-driver) v1.4.2, actively maintained.
- Keep `nano-interactive/ext-cassandra` as a fallback: the new driver needs PHP 8.3+ and CMake 3.30+.
- Enable the extension on Debian by building `apache/cassandra-cpp-driver` 2.17.1 from source when no package provides it (drops `!buster`, `!bullseye`, `!bookworm`, `!trixie`).
- Add `scripts/tests/cassandra`, working with both drivers.
- Tested locally on 12 image/PHP combinations, Alpine and Debian, PHP 7.4–8.5, including arm64.

### DIFF
--- a/data/special-requirements
+++ b/data/special-requirements
@@ -1,4 +1,4 @@
-cassandra !jessie !stretch !buster !bullseye !bookworm !trixie
+cassandra !jessie !stretch
 ddtrace !jessie
 ecma_intl !buster !bullseye !trixie !alpine3.22 !alpine3.23 !alpine3.24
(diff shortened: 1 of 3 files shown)


mlocati:
Thanks!
I've updated this PR in https://github.com/mlocati/docker-php-extension-installer/pull/1364

PR #1318, opened 2026-09-10
Update geos to latest. Adds php 8.5 support, breaks php <7 support.

pvandommelen:
Updates the geos extension to the latest commit which adds php 8.5 support, but drops support php < 7 versions.


mlocati:
Thanks! #1321 doesn't break PHP 5.x support

pvandommelen:
Even better. Thanks!

